PR #5, opened 2015-03-24
Adding Python 3.4 to Travis and tox config.

dhermes:
Had to remove `google-apitools[testing]` for Python 3 tests.

---------

**ORIGINALLY SAID**:

Though this adds Python 3.4 to Travis, it doesn't currently pass due to a syntax error:

```python
  File ".../apitools/apitools/base/py/__init__.py", line 5, in <module>
    from apitools.base.py.base_api import *
  File ".../apitools/apitools/base/py/base_api.py", line 17, in <module>
    from apitools.base.py import credentials_lib
  File ".../apitools/apitools/base/py/credentials_lib.py", line 15, in <module>
    import oauth2client.tools  # for flag declarations
  File ".../apitools/.tox/py34/lib/python3.4/site-packages/oauth2client/tools.py", line 242, in <module>
    from oauth2client.old_run import run
  File ".../apitools/.tox/py34/lib/python3.4/site-packages/oauth2client/old_run.py", line 25, in <module>
    import gflags
  File ".../apitools/.tox/py34/lib/python3.4/site-packages/gflags.py", line 1091
    except gflags_validators.Error, e:
                                  ^
SyntaxError: invalid syntax
```

It may just be that you don't want those `gflags` tests to run under Python 3?

### DIFF
--- a/apitools/base/py/transfer.py
+++ b/apitools/base/py/transfer.py
@@ -9,7 +9,6 @@ import io
 import json
 import mimetypes
 import os
-import StringIO
 import threading
 
 import six
@@ -632,7 +631,7 @@ class Upload(_Transfer):
 
     # encode the body: note that we can't use `as_string`, because
     # it plays games with `From ` lines.
-    fp = StringIO.StringIO()
+    fp = io.BytesIO()
     g = email_generator.Generator(fp, mangle_from_=False)
     g.flatten(msg_root, unixfrom=False)
     http_request.body = fp.getvalue()
--- a/apitools/base/py/transfer.py
+++ b/apitools/base/py/transfer.py
@@ -631,7 +631,7 @@ class Upload(_Transfer):
 
     # encode the body: note that we can't use `as_string`, because
     # it plays games with `From ` lines.
-    fp = io.BytesIO()
+    fp = six.StringIO()  # email: cStringIO in Py2 / io.StringIO in Py3.
     g = email_generator.Generator(fp, mangle_from_=False)
     g.flatten(msg_root, unixfrom=False)
     http_request.body = fp.getvalue()
--- a/apitools/base/py/transfer.py
+++ b/apitools/base/py/transfer.py
@@ -629,9 +629,11 @@ class Upload(_Transfer):
     msg.set_payload(self.stream.read())
     msg_root.attach(msg)
 
-    # encode the body: note that we can't use `as_string`, because
-    # it plays games with `From ` lines.
-    fp = six.StringIO()  # email: cStringIO in Py2 / io.StringIO in Py3.
+    # NOTE: We encode the body, but can't use `email.message.Message.as_string`
+    #       because it prepends `> ` to `From ` lines.
+    # NOTE: We must use six.StringIO() instead of io.StringIO() since the
+    #       `email` library uses cStringIO in Py2 and io.StringIO in Py3.
+    fp = six.StringIO()
     g = email_generator.Generator(fp, mangle_from_=False)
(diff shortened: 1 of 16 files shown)


dhermes:
**UPDATE**: @craigcitro I have all the tests passing in Python 3, but it required hacking the `gflags` source to update `except ExceptionType, exc:` statements to `except ExceptionType as exc:` and to update Python 2 `print` syntax to Python 3 `print` syntax.

craigcitro:
@dhermes woooo this looks awesome.

what happens if instead of tweaking gflags (which is complicated because it's hard to update that pypi package), we instead only install gflags in we're in py2? 
* at least in the case above, that would use `argparse` inside `oauth2client` instead. 
* i suspect other tests might fail, specifically in `gen/`, but we could disable those few.

dhermes:
@craigcitro Check out the latest commit.

apitools/base/py/transfer.py
```diff
@@ -632,7 +631,7 @@ def __ConfigureMultipartRequest(self, http_request):
 
     # encode the body: note that we can't use `as_string`, because
     # it plays games with `From ` lines.
-    fp = StringIO.StringIO()
+    fp = six.StringIO()  # email: cStringIO in Py2 / io.StringIO in Py3.
```
craigcitro: what's the comment here?

samples/storage_sample/uploads_test.py
```diff
@@ -37,8 +37,8 @@ def tearDown(self):
 
   def __ResetUpload(self, size, auto_transfer=True):
     self.__content = ''.join(
-        random.choice(string.letters) for _ in xrange(size))
-    self.__buffer = StringIO.StringIO(self.__content)
+        random.choice(string.ascii_letters) for _ in xrange(size))
```
craigcitro: do you want to get this `xrange` while you're here?

craigcitro:
LGTM -- just 1-2 questions and we're good to go.

apitools/base/py/transfer.py
dhermes: Sadly the `email` module is a PITA. It doesn't behave consistently, so we have to use `six.StringIO()` instead of `io.StringIO()`.

craigcitro: ah, that makes sense -- let's maybe tweak the comment to say "We use six.StringIO instead of io.StringIO for compatibility with the email module" or the like?

samples/storage_sample/uploads_test.py
dhermes: Done. Also got removed a docstring from a unit test case since nose picks up the wrong name in those cases.

apitools/base/py/transfer.py
dhermes: Yeah the comment is totally vague. Before I write a new one, what does the current comment above mean? I don't want to have two different thoughts string together as one.

craigcitro: oh -- so that one's actually an absurd corner case.

the `email` module has an `as_string` function that *should* be what we want to use here. however, by default it assumes you're sending an email (probably not a bad assumption, given the name), and helpfully turns any line starting with 

    From ...

into

    > From ...

we noticed this when some early uploads started picking up two extra bytes on the GCS side (and hence failing consistency checks). probably my favorite bug we hit.

dhermes: OK updated it and pushed over the existing clean-up commit.

dhermes:
@craigcitro All questions addressed and Travis complete. Shall we merge?

craigcitro:
I thought I already had ;)

Issue #4, opened 2015-03-24
Running tox locally on Ubuntu 14.04 fails


dhermes:
With a brand-new checkout, running `tox -e py27` fails with

```python
======================================================================
ERROR: testGeneration (apitools.gen.client_generation_test.ClientGenerationTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File ".../apitools/apitools/gen/client_generation_test.py", line 61, in testGeneration
    retcode = subprocess.call(args)
  File "/usr/lib/python2.7/subprocess.py", line 522, in call
    return Popen(*popenargs, **kwargs).wait()
  File "/usr/lib/python2.7/subprocess.py", line 710, in __init__
    errread, errwrite)
  File "/usr/lib/python2.7/subprocess.py", line 1327, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory
-------------------- >> begin captured logging << --------------------
root: INFO: Testing API drive.v2 with command line: gen_client --client_id=12345 --client_secret=67890 --discovery_url=drive.v2 --outdir=generated --overwrite client
--------------------- >> end captured logging << ---------------------
```

dhermes:
OK this passed on the second run, but it's still an issue for the first run. I suppose the focus of this bug should be on that instead.

Secondary question: Is there really onlt 61 unit tests for this whole codebase? Running

```
git ls-files | grep -e 'py$' | egrep -v test | egrep -v setup | xargs wc -l
```

yields 12981 lines of code. 61 unit tests?

dhermes:
Running

```
$ nosetests -v --collect-only
testAdditionalHeaders (apitools.base.py.base_api_test.BaseApiTest) ... ok
testIncludeEmptyFieldsClient (apitools.base.py.base_api_test.BaseApiTest) ... ok
testJsonResponse (apitools.base.py.base_api_test.BaseApiTest) ... ok
testNoCredentials (apitools.base.py.base_api_test.BaseApiTest) ... ok
testPathRemapping (apitools.base.py.base_api_test.BaseApiTest) ... ok
testPrettyPrintEncoding (apitools.base.py.base_api_test.BaseApiTest) ... ok
testQueryEncoding (apitools.base.py.base_api_test.BaseApiTest) ... ok
testQueryRemapping (apitools.base.py.base_api_test.BaseApiTest) ... ok
testUrlNormalization (apitools.base.py.base_api_test.BaseApiTest) ... ok
[...]

PR #9, opened 2015-03-31
Fix an issue with setup scripts in tox.ini.

craigcitro:
This tweaks `tox.ini` to install additional packages in such a way that we
don't run afoul of the issues in #4.

@tseaver do you have any sense for *why* this would fix the failures we're seeing in #4? this is far beyond my python-packaging-fu. in particular, danny found [this link](https://bitbucket.org/hpk42/tox/issue/77/cant-install-new-extras-for-target-package) with the fix, but I'm still confused as to why this particular fix works, whereas other attempts (including *not* calling `pip` from inside `tox.ini`) didn't.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -4,7 +4,7 @@ envlist = py26,py27,pypy,py33,py34
 [testenv]
 deps = nose
 commands =
-    pip install google-apitools[testing]
+    pip install -e .[testing] -q --log={envlogdir}/pip-extra-install.log
     nosetests
 
 [testenv:py33]


tseaver:
@craigcitro the first example uses the `[testing]` extras *from the PyPI version* (rather than the local checkout).

I think I would probably just punt on the DRY and inline the list into `tox.ini` -- anybody who is likely to care will be running `tox` anyhow.

tseaver:
The Travis 3.4 failure is likely due to hash order stuff:  pass `sort_keys=True` to `json.dumps()` to normalize.

craigcitro:
@tseaver i had thought the same thing about dropping the `pip` line, but @dhermes said it failed when he tried. that leaves me utterly confused. :wink: 

i'll fix the `sort_keys` thing as soon as i get a sec.

tseaver:
My take on a better fix is in #10.

PR #10, opened 2015-03-31
Testing neaten

tseaver:
DRY tox dependencies (don't depend on setuptools 'testing' extra for it).

 Speed up Travis builds via `sudo: false` (see: http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/ )

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: python
+sudo: false
 env:
   - TOX_ENV=py27
   - TOX_ENV=pypy
--- a/tox.ini
+++ b/tox.ini
@@ -1,46 +1,48 @@
 [tox]
 envlist = py26,py27,pypy,py33,py34
 
+[minimal]
+deps =
+    nose
+    mock
+    unittest2
+
 [testenv]
-deps = nose
+deps =
+    {[minimal]deps}
+    google-apputils
 commands =
-    pip install google-apitools[testing]
     nosetests
 
 [testenv:py33]
 basepython = python3.3
 deps =
-    mock
-    nose
-    unittest2
-commands = nosetests
+    {[minimal]deps}
+commands =
+    nosetests
 
 [testenv:py34]
 basepython = python3.4
 deps =
-    mock
-    nose
-    unittest2
-commands = nosetests
+    {[minimal]deps}
+commands =
+    nosetests
 
 [testenv:cover]
 basepython =
     python2.7
-commands =
-    nosetests --with-xunit --with-xcoverage --cover-package=apitools --nocapture --cover-erase --cover-tests --cover-branches
 deps =
-    google-apputils
-    mock
-    nose
-    unittest2
+    {[testenv]deps}
     coverage
     nosexcover
+commands =
+    nosetests --with-xunit --with-xcoverage --cover-package=apitools --nocapture --cover-erase --cover-tests --cover-branches
 
 [testenv:coveralls]
 basepython = {[testenv:cover]basepython}
-commands =
-    {[testenv:cover]commands}
-    coveralls
 deps =
     {[testenv:cover]deps}
+    coveralls
+commands =
+    {[testenv:cover]commands}
     coveralls


.travis.yml
```diff
@@ -1,4 +1,5 @@
 language: python
+sudo: false
```
dhermes: Why?

tox.ini
```diff
@@ -1,46 +1,48 @@
 [tox]
 envlist = py26,py27,pypy,py33,py34
 
+[minimal]
```
dhermes: Minimal doesn't seem to be used anywhere but in `[testenv]`.

dhermes:
This still fails to finish the install and does not fix #4 on my machine: https://gist.github.com/dhermes/f0c0650526c8d62f9230

In particular, check out the results of `ls -1 .tox/py27/bin/g*` after the first and second runs of `tox -e py27`. The failure is because `gen_client` doesn't get created when the environment is brand new

.travis.yml
craigcitro: makes builds faster (at least, puts us in a higher throughput queue), and travis is trying to push everyone [that direction anyway](http://blog.travis-ci.com/2015-03-31-docker-default-on-the-way/).

craigcitro:
@dhermes just to confirm: is this just a vanilla ubuntu 14.04 machine? if so i want to try and repro this in GCE, to understand what's going on.

either this is an issue in my `tox.ini`/`setup.py`, or it's a bug in `tox`. i'm quite wary about claiming the latter.

dhermes:
Yes vanilla Ubuntu. (Also I got `sudo: false` mixed up with `sudo: true`; we use `false` in `gcloud-python` and I'm the one that committed it.)

tox.ini
tseaver: Look at `[testenv:py33]` and `[testenv:py34]`.

craigcitro:
grr, so i can't repro this on a GCE instance. you're 14.04, with recent python/pip/tox?

tseaver:
Here is mine:

  https://gist.github.com/tseaver/cfbacaae0426c428f2e0

tox.ini
dhermes: My bad.

dhermes:
```
$ uname -a
Linux ubuntlaptop 3.13.0-48-generic #80-Ubuntu SMP Thu Mar 12 11:16:15 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux
$ cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=14.04
DISTRIB_CODENAME=trusty
DISTRIB_DESCRIPTION="Ubuntu 14.04.2 LTS"
$ tox --version
1.9.2 imported from /usr/local/lib/python2.7/dist-packages/tox/__init__.pyc
$ pip --version
pip 6.0.8 from /usr/local/lib/python2.7/dist-packages (python 2.7)
```

dhermes:
Why are we wrestling all this when the failing test needs to be pulled out into a system / integration test anyhow?

craigcitro:
@dhermes so i don't care about the test so much -- but the core issue is that apitools isn't getting installed correctly on your machine, right?

or is this *only* under `tox`?

dhermes:
@craigcitro It's only under `tox`. I just tried with a virtual env and had no issues:

```
$ virtualenv venv
...
$ source venv/bin/activate
...
(venv)$ which gen_client
# Nothing
(venv)$ pip install google-apitools[testing]
...  # No failures printed
(venv)$ which gen_client
...venv/bin/gen_client
```

craigcitro:
@dhermes so wait -- if you just add a new tox env with

    [testenv:confusion]
    basepython = python2.7
    deps = 
        mock
        nose
    commands = echo "hello"

and run `tox -e confusion`, the new env doesn't have `gen_client` at all?

dhermes:
`tox` doesn't like `echo`, so I replaced with `commands = python -c 'print "hello"'`.

Just as with `py27`, the first run fails to create `gen_client` and the second does so successfully:
https://gist.github.com/dhermes/6b42e9242c4f531ea643

craigcitro:
@dhermes does it say anything insightful if you do `tox -v -v`?

dhermes:
Nope. But it did point out that on the first run, the "install" gets logged to `.tox/py27/log/py27-2.log` and on the second run to `.tox/log/tox-0.log`. The first log is very different, it runs

```
pip install .tox/dist/google-apitools-0.4.1.zip
```

while the second runs

```
python setup.py sdist ...
```

to recreate the zip.

cherba:
Is anyone still pursuing this PR, do you mind if we close it?

tseaver:
I'm not pushing for it.

PR #11, opened 2015-03-31
Fixing non-deterministic dict. failure in extra_types_test.

dhermes:
Failure occurs when comparing encoded strings via `json.dumps` and `apitools.base.py.encoding.MessageToJson`.

The `MessageToJson` function has non-deterministic results, as does `json.dumps`.

@craigcitro This fixes the failure in #9.

### DIFF
--- a/apitools/base/py/extra_types_test.py
+++ b/apitools/base/py/extra_types_test.py
@@ -125,12 +125,12 @@ class ExtraTypesTest(unittest2.TestCase):
             datetime.date(1980, 10, 24),
             datetime.date(1981, 1, 19),
             ])
-    json_msg = json.dumps({
-        'start_date': '1752-09-09', 'all_dates': [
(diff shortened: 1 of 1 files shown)


dhermes:
@craigcitro Are you not concerned that `encoding.MessageToJson` has non-deterministic behavior?

craigcitro:
it's something to clean up, but we only really use JSON encoding for sending things over the wire (and the server doesn't care about key order). unless i'm missing something, this should only be an issue for tests that use literal json strings, right?

i was planning to file an issue later, but i'll do it now.

dhermes:
Yeah I wasn't sure it mattered, but figured it might because this test failure.

craigcitro:
i have a record/replay framework on top of apitools internally that i'm planning to push out at some point; *that* would likely be more fragile on this front. ;)

dhermes:
Nice!

PR #13, opened 2015-04-08
Upgrading oauth2client.tools.run to run_flow.

dhermes:
@craigcitro I assume this won't be sufficient / you need some legacy support for `gflags`.

NOTE: I reference the warning for the out-of-date method in GoogleCloudPlatform/gcloud-python#755

---

I tested this with 

```python
# gen_client --discovery_url=storage.v1 --nogenerate_cli --outdir=storage client
from storage import storage_v1_client as storage_client
from storage import storage_v1_messages as storage_messages
client = storage_client.StorageV1()
```

and it works just fine.

### DIFF
--- a/apitools/base/py/credentials_lib.py
+++ b/apitools/base/py/credentials_lib.py
@@ -2,6 +2,7 @@
 """Common credentials classes and constructors."""
 from __future__ import print_function
 
+import argparse
 import datetime
 import json
 import os
@@ -13,7 +14,7 @@ import oauth2client.gce
 import oauth2client.locked_file
 import oauth2client.multistore_file
 import oauth2client.service_account
-import oauth2client.tools  # for flag declarations
+from oauth2client import tools  # for gflags declarations
 from six.moves import http_client
 from six.moves import urllib
 
@@ -384,6 +385,21 @@ class GaeAssertionCredentials(oauth2client.client.AssertionCredentials):
     self.access_token = token
 
 
+def _GetRunFlowFlags():
+  parser = argparse.ArgumentParser(parents=[tools.argparser])
+  # Get command line argparse flags.
+  flags = parser.parse_args()
+
+  # Allow `gflags` and `argparse` to be used side-by-side.
+  if hasattr(FLAGS, 'auth_host_name'):
+    flags.auth_host_name = FLAGS.auth_host_name
+  if hasattr(FLAGS, 'auth_host_port'):
+    flags.auth_host_port = FLAGS.auth_host_port
+  if hasattr(FLAGS, 'auth_local_webserver'):
+    flags.noauth_local_webserver = (not FLAGS.auth_local_webserver)
+  return flags
+
+
 # TODO(craigcitro): Switch this from taking a path to taking a stream.
 def CredentialsFromFile(path, client_info):
   """Read credentials from a file."""
@@ -403,9 +419,8 @@ def CredentialsFromFile(path, client_info):
       # retry loop, they can ^C.
       try:
         flow = oauth2client.client.OAuth2WebServerFlow(**client_info)
-        # We delay this import because it's rarely needed and takes a long time.
-        from oauth2client import tools
-        credentials = tools.run(flow, credential_store)
+        flags = _GetRunFlowFlags()
+        credentials = tools.run_flow(flow, credential_store, flags)
         break
       except (oauth2client.client.FlowExchangeError, SystemExit) as e:
         # Here SystemExit is "no credential at all", and the
--- a/apitools/base/py/credentials_lib_test.py
+++ b/apitools/base/py/credentials_lib_test.py
@@ -63,5 +63,38 @@ class CredentialsLibTest(unittest2.TestCase):
     self._GetServiceCreds(service_account_name='my_service_account')
 
 
+class TestGetRunFlowFlags(unittest2.TestCase):
+
+  def setUp(self):
+    self._flags_actual = credentials_lib.FLAGS
+
+  def tearDown(self):
+    credentials_lib.FLAGS = self._flags_actual
+
+  def test_with_gflags(self):
+    HOST = object()
+    PORT = object()
+
+    class MockFlags(object):
+      auth_host_name = HOST
+      auth_host_port = PORT
+      auth_local_webserver = False
+
+    credentials_lib.FLAGS = MockFlags
+    flags = credentials_lib._GetRunFlowFlags()
+    self.assertEqual(flags.auth_host_name, HOST)
+    self.assertEqual(flags.auth_host_port, PORT)
+    self.assertEqual(flags.logging_level, 'ERROR')
+    self.assertEqual(flags.noauth_local_webserver, True)
+
+  def test_without_gflags(self):
+    credentials_lib.FLAGS = None
+    flags = credentials_lib._GetRunFlowFlags()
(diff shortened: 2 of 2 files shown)


craigcitro:
yeah, i can't just drop `gflags` -- it'll break all internal use of this code.

the only option here is the (awkward) "try `gflags` and `tools.run`, then fall back to `argparse` and `tools.run_flow`". i almost think we should just roll that into `oauth2client`, but i haven't really thought it through.

dhermes:
Nope, there is a "better" option that let's you move off of `gflags` at your own pace:

```python
flags = argparse.Namespace(auth_host_name=FLAGS.auth_host_name,
                           auth_host_port=FLAGS.auth_host_port,
                           logging_level='ERROR',
                           noauth_local_webserver=not FLAGS.auth_local_webserver)
```

craigcitro:
ah, so you're saying we can drop `tools.run` for `tools.run_flow` -- we just need to conditionally load from `flags.FLAGS` or a passed-in namespace. sgtm.

dhermes:
@craigcitro PTAL. I made the changes to incorporate `gflags` and added a test for the method I added.

apitools/base/py/credentials_lib.py
```diff
@@ -13,7 +14,9 @@
 import oauth2client.locked_file
 import oauth2client.multistore_file
 import oauth2client.service_account
-import oauth2client.tools  # for flag declarations
+# The `tools` import is rarely needed, but takes a long time.
```
dhermes: @craigcitro I ported this comment from the actual changed section below. It seems to be out-of-date (i.e. the slow import is just `oauth2client.crypt` or anything that imports it) but I left it in just in case.

apitools/base/py/credentials_lib.py
```diff
@@ -384,6 +387,21 @@ def _refresh(self, _):
     self.access_token = token
 
 
+def _GetRunFlowFlags():
+  if FLAGS is None:
+    parser = argparse.ArgumentParser(parents=[tools.argparser])
+    # We pass [] to just use the defaults from tools.argparser instead of
+    # using command line flags.
+    return parser.parse_args([])
```
dhermes: @craigcitro Note that this disables using the arguments (i.e. `sys.argv`). Do you want to allow command line flags for both `argparse` and `gflags`? It's a simple change to `parser.parse_args()` (no empty list passed) if we want to support `argparse` from the command line as well.

apitools/base/py/credentials_lib.py
```diff
@@ -384,6 +387,21 @@ def _refresh(self, _):
     self.access_token = token
 
 
+def _GetRunFlowFlags():
+  if FLAGS is None:
```
dhermes: @craigcitro This differs from your approach below: `if hasattr(FLAGS, 'auth_local_webserver'):`

Is there a reason you didn't just check `FLAGS is not None`?

apitools/base/py/credentials_lib.py
craigcitro: yeah, i vaguely remember running into this and not digging much. you can probably drop this whole comment now (since we use `tools` for the arg parser).

apitools/base/py/credentials_lib.py
craigcitro: yeah, let's pick up `sys.argv` here, as one step towards using `argparse`.

apitools/base/py/credentials_lib.py
craigcitro: yes -- because of the way `gflags` works, the `auth_local_webserver` flag only gets declared if the `tools` module was loaded. so at some point i had an internal build where `FLAGS` was available, but `FLAGS.auth_local_webserver` threw an `AttributeError`.

i found and fixed the root cause in that case, but it seemed safe enough to leave the code as extra cautious, in the event it ever came back.

apitools/base/py/credentials_lib.py
dhermes: It's still [needed][1] for the `gflags` declarations because it imports `old_run`. The "long time" thing is what I was getting hung up on.

[1]: https://github.com/google/oauth2client/blob/0a6241c792f0c833f2f176292b0c7ea5623eabfd/oauth2client/old_run.py#L36-L47

apitools/base/py/credentials_lib.py
dhermes: OK. Should I take that approach with this method or `FLAGS is not None` is sufficient?

apitools/base/py/credentials_lib.py
```diff
@@ -384,6 +387,21 @@ def _refresh(self, _):
...
+    parser = argparse.ArgumentParser(parents=[tools.argparser])
+    # We pass [] to just use the defaults from tools.argparser instead of
+    # using command line flags.
+    return parser.parse_args([])
+  else:
+    return argparse.Namespace(
+        auth_host_name=FLAGS.auth_host_name,
+        auth_host_port=FLAGS.auth_host_port,
+        logging_level='ERROR',
+        noauth_local_webserver=not FLAGS.auth_local_webserver,
```
craigcitro: can you parenthesize the RHS (just to make it easier to read)?

apitools/base/py/credentials_lib_test.py
```diff
@@ -63,5 +63,38 @@ def testGceServiceAccounts(self):
     self._GetServiceCreds(service_account_name='my_service_account')
 
 
+class Test__GetRunFlowFlags(unittest2.TestCase):
```
craigcitro: you can just call it `TestGetRunFlowFlags` -- easier to read and unlikely to be confusing.

craigcitro:
LGTM, one or two little comments/questions

apitools/base/py/credentials_lib.py
dhermes: Will do, but it is somewhat un-Pythonic to add parens when not needed.

dhermes:
Yup. I've got 4 changes to make and one remaining questions for you (about `FLAGS is None` vs. checking attrs)

apitools/base/py/credentials_lib.py
craigcitro: given that i voted for allowing a mix of `flags` and `argparse` below, probably better safe than sorry.

feel free to leave a `TODO` for me to clean this up when i drop `flags` completely.

apitools/base/py/credentials_lib.py
craigcitro: i agree in general, but i think legibility is the bigger concern.

dhermes:
@craigcitro I pushed over the original commit with the changes. Main update is just checking for each attribute on `FLAGS`. PTAL.

apitools/base/py/credentials_lib.py
```diff
@@ -13,7 +14,8 @@
 import oauth2client.locked_file
 import oauth2client.multistore_file
 import oauth2client.service_account
-import oauth2client.tools  # for flag declarations
+# It is needed for gflags declarations.
```
craigcitro: nit: `It` is ambiguous (is it the next declaration or the previous one?). maybe either name the module or keep it on the same line?

craigcitro:
looks great -- one really nitpicky comment and then we're good to go. :+1:

dhermes:
Yeah I noticed the ambiguous `it` too. Removing a line is not always sufficient in a human language :)

Addressed and pushed over the original commit.

Issue #18, opened 2015-04-16
Upload.__StreamMedia breaks on non-seekable streams


mwitkow-io:
I'd like to be able to use non-seekable streams, such as `werkzeug.wsgi.LimitedStream` that is used in Flask for streaming request and respone data. This is how we do streaming to/from Google Cloud Storage. However, the `Upload.__StreamMedia` calls `stream.seek()` to verify completion:

```
  File "site-packages/gcloud/storage/blob.py", line 350, in upload_from_file
    finish_callback=lambda *args: None)
  File "site-packages/apitools/base/py/transfer.py", line 807, in StreamInChunks
    additional_headers=additional_headers)
  File "site-packages/apitools/base/py/transfer.py", line 773, in __StreamMedia
    self.stream.seek(0, os.SEEK_END)
AttributeError: 'LimitedStream' object has no attribute 'seek'
```

There's even a TODO:
```
# TODO(craigcitro): Decide how to handle errors in the
      # non-seekable case.
```

PR #19, opened 2015-04-16
Fix uploads for non-seekable streams

mwitkow-io:
Fixes https://github.com/google/apitools/issues/18


mwitkow-io:
I signed it!

mwitkow-io:
Or at least our company did "Improbable Worlds Ltd"

craigcitro:
this looks good -- just want to ask @thobrla to confirm that this is unlikely to quietly break anything for gsutil?

thobrla:
Looks good.

Note: we use seekable() elsewhere but that only applies to IOBase objects, which is not applicable for upload streams with type 'file'.

craigcitro:
awesome -- thanks @mwitkow-io!

PR #22, opened 2015-04-30
Change callback defaults for transfers.

craigcitro:
This makes several changes to the way we use callbacks in media transfers:
 * allow callbacks to be attached to a transfer, not just be provided as an
 argument to streaming functions,
 * switch the defaults to "don't print",
 * expose the "print progress" defaults as public functions instead of
 protected static methods, and
 * teach the generated CLIs to print by default.

PTAL @thobrla and @dhermes 

@thobrla -- if this looks good, I can go in and remove all the "null callback" business in gsutil after the next apitools release.


thobrla:
LGTM

PR #23, opened 2015-05-04
Fix credentials_lib tests to not rely on sys.argv.

craigcitro:
Previously, some tests in credentials_lib were implicitly reading argv (via
`argparse`); this cleans them up to have an explicit argv passed in.

PTAL @dhermes


dhermes:
LGTM

PR #24, opened 2015-05-05
Validate user-provided MIME types on uploads.

craigcitro:
Currently, we'll allow any string as the MIME type -- but valid MIME types
must contain a `/`. The code implicitly assumes this, so we might as well
check it and raise a comprehensible error.

PTAL @thobrla


thobrla:
LGTM

PR #29, opened 2015-05-20
Conditionally import argparse in credentials_lib.

craigcitro:
This is a bit odd, but avoids a much nastier workaround in gsutil.

PTAL @thobrla


thobrla:
LGTM

PR #30, opened 2015-05-22
Restore chunked download support.

craigcitro:
We lost support for chunked downloads in a merge; this PR restores chunked
download support, and adds tests.

In particular, since gsutil does its own chunking, we do one-shot downloads in
the case that `download.chunksize == 0`.

PTAL @thobrla -- i'm sure there are some corner cases I'm not yet thinking of.

@charlesccychen I think this should get you fixed up.

### DIFF
--- a/apitools/base/py/transfer.py
+++ b/apitools/base/py/transfer.py
@@ -298,6 +298,8 @@ class Download(_Transfer):
             http_request.url = client.FinalizeTransferUrl(http_request.url)
         url = http_request.url
         if self.auto_transfer:
+            end_byte = self.__ComputeEndByte(0)
+            self.__SetRangeHeader(http_request, 0, end_byte)
             response = http_wrapper.MakeRequest(
                 self.bytes_http or http, http_request)
             if response.status_code not in self._ACCEPTABLE_STATUSES:
@@ -339,14 +341,42 @@ class Download(_Transfer):
         else:
             request.headers['range'] = 'bytes=%d-%d' % (start, end)
 
-    def __GetChunk(self, start, end=None, additional_headers=None):
+    def __ComputeEndByte(self, start, end=None, use_chunks=True):
+        """Compute the last byte to fetch for this request.
+
+        This is all based on the HTTP spec for Range and
+        Content-Range.
+
+        Note that this is potentially confusing in several ways:
+          * the value for the last byte is 0-based, eg "fetch 10 bytes
+            from the beginning" would return 9 here.
+          * if we have no information about size, and don't want to
+            use the chunksize, we'll return None.
+        See the tests for more examples.
+
+        Args:
+          start: byte to start at.
+          end: (int or None, default: None) Suggested last byte.
+          use_chunks: (bool, default: True) If False, ignore self.chunksize.
+
+        Returns:
+          Last byte to use in a Range header, or None.
+
+        """
+        end_byte = end
+        if use_chunks:
+            alternate = start + self.chunksize - 1
+            end_byte = min(end_byte, alternate) if end_byte else alternate
+        if self.total_size:
+            alternate = self.total_size - 1
+            end_byte = min(end_byte, alternate) if end_byte else alternate
+        return end_byte
+
+    def __GetChunk(self, start, end, additional_headers=None):
         """Retrieve a chunk, and return the full response."""
         self.EnsureInitialized()
-        end_byte = end
-        if self.total_size and end:
-            end_byte = min(end, self.total_size)
         request = http_wrapper.Request(url=self.url)
-        self.__SetRangeHeader(request, start, end=end_byte)
+        self.__SetRangeHeader(request, start, end=end)
         if additional_headers is not None:
             request.headers.update(additional_headers)
         return http_wrapper.MakeRequest(
@@ -378,7 +408,8 @@ class Download(_Transfer):
             self.stream.write('')
         return response
 
-    def GetRange(self, start, end=None, additional_headers=None):
+    def GetRange(self, start, end=None, additional_headers=None,
+                 use_chunks=True):
         """Retrieve a given byte range from this download, inclusive.
 
         Range must be of one of these three forms:
@@ -394,6 +425,8 @@ class Download(_Transfer):
           end: (int, optional) Where to stop fetching bytes. (See above.)
           additional_headers: (bool, optional) Any additional headers to
               pass with the request.
+          use_chunks: (bool, default: True) If False, ignore self.chunksize
+              and fetch this range in a single request.
 
         Returns:
           None. Streams bytes into self.stream.
@@ -406,7 +439,9 @@ class Download(_Transfer):
         else:
             progress = start
         while not progress_end_normalized or progress < end:
-            response = self.__GetChunk(progress, end=end,
+            end_byte = self.__ComputeEndByte(progress, end=end,
+                                             use_chunks=use_chunks)
+            response = self.__GetChunk(progress, end_byte,
                                        additional_headers=additional_headers)
             if not progress_end_normalized:
                 self.__SetTotal(response.info)
@@ -420,7 +455,28 @@ class Download(_Transfer):
 
     def StreamInChunks(self, callback=None, finish_callback=None,
                        additional_headers=None):
-        """Stream the entire download."""
+        """Stream the entire download in chunks."""
+        self.StreamMedia(callback=callback, finish_callback=finish_callback,
+                         additional_headers=additional_headers,
+                         use_chunks=True)
+
+    def StreamMedia(self, callback=None, finish_callback=None,
+                    additional_headers=None, use_chunks=True):
+        """Stream the entire download.
+
+        Args:
+          callback: (default: None) Callback to call as each chunk is
+              completed.
+          finish_callback: (default: None) Callback to call when the
+              download is complete.
+          additional_headers: (default: None) Additional headers to
+              include in fetching bytes.
+          use_chunks: (bool, default: True) If False, ignore self.chunksize
+              and stream this download in a single request.
+
+        Returns:
+            None. Streams bytes into self.stream.
+        """
         callback = callback or self.progress_callback
         finish_callback = finish_callback or self.finish_callback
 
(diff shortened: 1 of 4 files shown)


apitools/base/py/transfer.py
```diff
@@ -344,7 +344,9 @@ def __GetChunk(self, start, end=None, additional_headers=None):
         self.EnsureInitialized()
         end_byte = end
         if self.total_size and end:
-            end_byte = min(end, self.total_size)
+            end_byte = min(end_byte, self.total_size)
+        # if self.chunksize > 0:
```
thobrla: I assume this is meant to be uncommented.

thobrla: In the _Transfer base class, we accept a chunksize kwarg and use it like so:
self.chunksize = chunksize or 1048576

Which means a chunksize of 0 as a kwarg won't be respected.

apitools/base/py/transfer_test.py
```diff
@@ -12,6 +14,79 @@
...
+            self.assertEqual(base_url, make_request.call_args[0][1].url)
+            download_stream.seek(0)
+            self.assertEqual(string.ascii_lowercase * 2,
+                             download_stream.getvalue())
+
+    def testChunkedDownload(self):
+        bytes_http = object()
+        http = object()
+        download_stream = six.StringIO()
+        download = transfer.Download.FromStream(
```
thobrla: This download setup looks identical to the non-chunked one, so why would we expect different behavior from the http request?

apitools/base/py/transfer.py
```diff
@@ -344,7 +344,9 @@ def __GetChunk(self, start, end=None, additional_headers=None):
...
+        #     end_byte = min(end_byte, start + self.chunksize)
```
thobrla: I realize that this may also have implications for GetRange, but I like Upload's explicitly named parameter better than a sentinel value of 0 chunksize, precisely to avoid the kinds of issues I mentioned above.

Don't feel strongly about it, though.

apitools/base/py/transfer.py
craigcitro: yargh, good call.

apitools/base/py/transfer.py
craigcitro: cleaned up.

apitools/base/py/transfer_test.py
craigcitro: mocking at 1AM != good idea. :grinning:

craigcitro:
PTAL

apitools/base/py/transfer.py
```diff
@@ -420,7 +432,29 @@ def GetRange(self, start, end=None, additional_headers=None):
...
     def StreamInChunks(self, callback=None, finish_callback=None,
                        additional_headers=None):
-        """Stream the entire download."""
+        """Stream the entire download in chunks."""
+        self.StreamMedia(callback=callback, finish_callback=finish_callback,
+                         additional_headers=additional_headers,
+                         ignore_chunksize=False)
+
+    def StreamMedia(self, callback=None, finish_callback=None,
+                    additional_headers=None, ignore_chunksize=False):
```
thobrla: Here and elsewhere, use_chunks = true for consistency with Upload?

apitools/base/py/transfer.py
```diff
@@ -378,7 +386,8 @@ def __ProcessResponse(self, response):
             self.stream.write('')
         return response
 
-    def GetRange(self, start, end=None, additional_headers=None):
+    def GetRange(self, start, end=None, additional_headers=None,
+                 ignore_chunksize=False):
```
thobrla: This will require a gsutil revision (and thus an apitools release), which is fine.

apitools/base/py/transfer.py
```diff
@@ -339,12 +341,18 @@ def __SetRangeHeader(self, request, start, end=None):
...
 
-    def __GetChunk(self, start, end=None, additional_headers=None):
+    def __GetChunk(self, start, end=None, additional_headers=None,
+                   ignore_chunksize=False):
         """Retrieve a chunk, and return the full response."""
         self.EnsureInitialized()
         end_byte = end
-        if self.total_size and end:
-            end_byte = min(end, self.total_size)
+        if not ignore_chunksize:
```
thobrla: It feels a bit strange to have a function named GetChunk care about chunking.  Why not push the logic around "end" into a helper function that's called by StreamMedia + GetRange?

apitools/base/py/transfer.py
```diff
@@ -420,7 +432,29 @@ def GetRange(self, start, end=None, additional_headers=None):
...
+
+        Args:
+          callback: (default: None) Callback to call as each chunk is
+              completed.
+          finish_callback: (default: None) Callback to call when the
+              download is complete.
+          additional_headers: (default: None) Additional headers to
+              include in fetching bytes.
+          ignore_chunksize: (bool, default: False) If True, ignore
+              self.chunksize and attempt to stream this download in
```
thobrla: This is at odds with the function docstring.

apitools/base/py/transfer.py
craigcitro: hah ... done.

apitools/base/py/transfer.py
craigcitro: yeah -- it was really an accidental change in the default apitools behavior.

apitools/base/py/transfer.py
craigcitro: my thought was that i didn't want anyone to accidentally skip this logic (say if there were another method added).

i don't mind the naming bit -- it's figuring out *what* chunk to get. :grin: but i'm pulling it out to add tests.

charlesccychen:
LGTM

apitools/base/py/transfer.py
craigcitro: good call.

craigcitro:
PTAL

apitools/base/py/transfer.py
```diff
@@ -339,14 +341,39 @@ def __SetRangeHeader(self, request, start, end=None):
         else:
             request.headers['range'] = 'bytes=%d-%d' % (start, end)
 
-    def __GetChunk(self, start, end=None, additional_headers=None):
+    def __ComputeEndByte(self, start, end=None, use_chunks=True):
+        """Compute the last byte to fetch for this request.
+
+        Note that this is potentially confusing in several ways:
+          * the value for the last byte is 0-based, eg "fetch 10 bytes
```
thobrla: Nit: this is per the HTTP spec for Range/Content-Range, may be useful to relate that.

apitools/base/py/transfer.py
```diff
@@ -298,6 +298,8 @@ def InitializeDownload(self, http_request, http=None, client=None):
             http_request.url = client.FinalizeTransferUrl(http_request.url)
         url = http_request.url
         if self.auto_transfer:
+            end_byte = min(self.total_size or 0, self.chunksize) - 1
```
thobrla: Why not compute this with __ComputeEndByte?

apitools/base/py/transfer.py
craigcitro: done.

apitools/base/py/transfer.py
craigcitro: i don't believe in code reuse.

KIDDING! i forgot. done.

apitools/base/py/transfer.py
```diff
@@ -339,14 +341,39 @@ def __SetRangeHeader(self, request, start, end=None):
...
+        Returns:
+          Last byte to use in a Range header, or None.
+        """
+        end_byte = end
+        if use_chunks:
+            alternate = start + self.chunksize - 1
+            end_byte = min(end_byte, alternate) if end_byte else alternate
+        if self.total_size:
+            alternate = self.total_size - 1
+            end_byte = min(end_byte, alternate) if end_byte else alternate
```
thobrla: I don't think this is harmful, but it is a semantic change where __GetChunk used to set a range header of:
bytes=0-

and now sets
bytes=0-22   # example assuming self.total_size == 23

in the case where use_chunks=False and bool(self.total_size)

apitools/base/py/transfer.py
```diff
@@ -339,14 +341,39 @@ def __SetRangeHeader(self, request, start, end=None):
...
+
+        See the tests for more examples.
+
+        Args:
+          start: byte to start at.
+          end: (int or None, default: None) Suggested last byte.
+          use_chunks: (bool, default: True) If False, ignore self.chunksize.
+
+        Returns:
+          Last byte to use in a Range header, or None.
```
thobrla: This has some potentially strange interaction with negative start ranges.  Is it intended for use in that case?

apitools/base/py/transfer.py
craigcitro: yeah, i thought about that when i was writing the tests; i also can't think of a case where it's harmful. 

well, actually, that's not true: if you had *mis-specified* the `total_size`, you'll now get an error trying to download instead of silently getting too many bytes. it hink that's better?

apitools/base/py/transfer.py
craigcitro: so at least at this layer, the HTTP spec doesn't support negative start indices in the range.

in the case of something like `GetRange(-10)`, we normalize first (see `__NormalizeStartEnd`).

craigcitro:
and i've confirmed that `gsutil test -p 100` passes with this code patched in.

apitools/base/py/transfer_test.py
```diff
@@ -12,6 +14,146 @@
...
+               'response_range ({1})').format(
+                   request_range, response_range)
+        self.assertTrue(response_range.startswith(request_range), msg=msg)
+
+    def testComputeEndByte(self):
+        total_size = 100
+        chunksize = 10
+        download = transfer.Download.FromStream(
+            six.StringIO(), chunksize=chunksize, total_size=total_size)
+        self.assertEqual(9, download._Download__ComputeEndByte(0, end=50))
```
thobrla: Nit: chunksize - 1

craigcitro: touche!

apitools/base/py/transfer_test.py
```diff
@@ -12,6 +14,146 @@
...
+                    content=string.ascii_lowercase,
+                    request_url='https://part.one/',
+                )
+            elif url == 'https://part.two/':
+                return http_wrapper.Response(
+                    info={
+                        'content-range': 'bytes 26-51/52',
+                        'status': http_client.OK,
+                    },
+                    content=string.ascii_lowercase,
```
thobrla: It would probably be good to differentiate the bytes in this case (so that the second response is different in content than the first)

craigcitro: good call, done.

thobrla:
LGTM

PR #36, opened 2015-06-24
Fixed off by one error in GetRange.

brandonsalmon22:
I believe this bug made it impossible to download ranges of size one (start_byte == end_byte).

### DIFF
--- a/apitools/base/py/transfer.py
+++ b/apitools/base/py/transfer.py
@@ -366,10 +366,18 @@ class Download(_Transfer):
         end_byte = end
         if use_chunks:
             alternate = start + self.chunksize - 1
-            end_byte = min(end_byte, alternate) if end_byte else alternate
+            if end_byte is not None:
+                end_byte = min(end_byte, alternate)
+            else:
+                end_byte = alternate
+
         if self.total_size:
             alternate = self.total_size - 1
-            end_byte = min(end_byte, alternate) if end_byte else alternate
+            if end_byte is not None:
+                end_byte = min(end_byte, alternate)
+            else:
+                end_byte = alternate
+
         return end_byte
 
     def __GetChunk(self, start, end, additional_headers=None):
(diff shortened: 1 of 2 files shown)


apitools/base/py/transfer.py
```diff
@@ -366,10 +366,12 @@ def __ComputeEndByte(self, start, end=None, use_chunks=True):
         end_byte = end
         if use_chunks:
             alternate = start + self.chunksize - 1
-            end_byte = min(end_byte, alternate) if end_byte else alternate
+            end_byte = (min(end_byte, alternate) if end_byte is not None
```
craigcitro: probably just make this an `if`/`else`. (line wrapping a ternary operator is hard to read.)

apitools/base/py/transfer.py
```diff
@@ -438,7 +440,7 @@ def GetRange(self, start, end=None, additional_headers=None,
             progress_end_normalized = True
         else:
             progress = start
-        while not progress_end_normalized or progress < end:
+        while not progress_end_normalized or progress < end + 1:
```
craigcitro: two things: 

1. let's just do `progress <= end`.
2. this seems like it's an off-by-one error in all cases. @thobrla have we had any other reports of an extra byte being returned? or were you calling this in gsutil as though it didn't include `end`?

thobrla: This is a bug (can repro with gsutil cat -r 1-1 gs://bucket/object), but GetRange with end < size is only used for the cat command with the range flag.  Otherwise, we're always downloading the object all the way to the end, so we won't hit this.

craigcitro:
looks good!

are there some reasonable tests we could throw in?

thobrla:
Seems like we should add a test in transfer_test.py.

Once this goes in, we can add a gsutil cat test as well.

brandonsalmon22:
I don't see any easy way to test it in transfer_test.py. Suggestions? The difficulty comes from GetRange not being directly tested in transfer_test.py.

Adding tests for gsutil cat will be simple, on the other hand.

craigcitro:
you could mock out `__GetChunk`, but that might get messy. maybe add some `__ComputeEndByte` tests and test `gsutil cat`?

thobrla:
Brandon and I discussed and I think we found a way to test in apitools using mocks.

PR #37, opened 2015-07-20
Add a `pip_package` command to gen_client.

craigcitro:
This is the most basic form of "generate a PyPI-viable package for a given
API". There's still some work to be done (namely around letting the caller
thread in version numbers), but this gets the job done in the simplest case.

PTAL @silviulica


apitools/gen/gen_client_lib.py
```diff
@@ -164,6 +170,73 @@ def WriteInit(self, out):
...
         printer('__path__ = pkgutil.extend_path(__path__, __name__)')
 
+    def WriteIntermediateInit(self, out):
+        """Write a simple __init__.py for an intermediate directory."""
+        printer = self._GetPrinter(out)
+        printer('#!/usr/bin/env python')
+        printer('"""Shared __init__.py for apitools."""')
+        printer()
+        printer('from pkgutil import extend_path')
+        printer('__path__ = extend_path(__path__, __name__)')
```
silviulica: Nice! This is the "let's not step on each others toes" code. Right?

apitools/gen/gen_client_lib.py
```diff
@@ -164,6 +170,73 @@ def WriteInit(self, out):
...
+        printer('# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either '
+                'express or implied.')
+        printer('# See the License for the specific language governing '
+                'permissions and')
+        printer('# limitations under the License.')
+        printer()
+        printer('import setuptools')
+        printer('REQUIREMENTS = [')
+        with printer.Indent(indent='    '):
+            # TODO(craigcitro): Have this track apitools' version.
```
silviulica: It is probably ok as long as it past the version that supports building this packages.

apitools/gen/gen_client.py
```diff
@@ -211,6 +235,33 @@ def Run(self, _):
...
+        package = discovery_doc['name']
+        original_outdir = os.path.expanduser(FLAGS.outdir)
+        FLAGS.outdir = os.path.join(
+            FLAGS.outdir, 'apitools/clients/%s' % package)
+        FLAGS.root_package = 'apitools.clients.%s' % package
+        FLAGS.generate_cli = False
+        codegen = _GetCodegenFromFlags()
+        if codegen is None:
+            logging.error('Failed to create codegen, exiting.')
+            return 128
```
silviulica: you want a -1 in 8 bits here?

silviulica:
Looks good Craig! Thanks for doing it.
The version stuff should we thread it into the namespaces (i.e. apitools/clients/storage_v1) or should we keep it only behind the scenes in dependencies? In a way I like more the explicit version in the name.

apitools/gen/gen_client_lib.py
craigcitro: yep! shamelessly stolen from [the docs](https://docs.python.org/2/library/pkgutil.html#pkgutil.extend_path). :grin:

apitools/gen/gen_client_lib.py
craigcitro: yeah, i think it's *generally* going to be safe -- but easy enough to fix at some point.

apitools/gen/gen_client.py
craigcitro: this was actually a copy-paste from above, and was probably a bad idea where it appears above. ;) (i needed a safe exit code to use for catching certain errors in tests.)

switched to 1.

craigcitro:
thanks silviu! submitting.

PR #39, opened 2015-07-26
Fix a bug in upload support, and do some cleanup.

craigcitro:
A recent cleanup introduced a NameError in a corner case; this was covered by
integration tests (yay!) that aren't run by default (boo!).

We clean this up (and fix a revealed corner-case bug), and clean up the tests
to run independent of the directory they're invoked from. Last, we add a tox
env (still for testing purposes) that computes coverage for the storage
integration tests; we're currently at 75%, covering most of the relevant code
in transfer.py.

PTAL @brandonsalmon22 @thobrla


thobrla:
LGTM

PR #40, opened 2015-07-26
Minor code cleanup in credentials_lib.

craigcitro:
PTAL @thobrla


apitools/base/py/credentials_lib.py
```diff
@@ -171,17 +182,15 @@ def __init__(self, scopes=None, service_account_name='default', **kwds):
...
         cached_scopes = None
-        if 'cache_filename' in kwds:
-            cache_filename = kwds['cache_filename']
+        cache_filename = kwds.get('cache_filename')
+        if cache_filename:
             cached_scopes = self._CheckCacheFileForMatch(
                 cache_filename, scopes)
 
-        scopes = cached_scopes or self._ScopesFromMetadataServer(scopes)
+        scopes = (scopes or cached_scopes or
```
thobrla: Why the change here?  If the metadata server doesn't support the scopes that we're trying to initialize with,  I think this should fail.

craigcitro: so note that before this change, *if* `cache_filename` was provided, and there were cached scopes, then we totally ignored the `scopes` argument provided by the user.

in particular, it means that it's possible the user would get back a token for different scopes than they requested.

thobrla: That's true, but the scopes would always be a superset of what was requested if we hit the cache file, because we check that the scopes the user requested exist in the cached scopes.

I guess I'd be okay calling a reduction in requested scopes a cache miss, but if we get a cache miss, it seems like we should always query the metadata server to ensure that it actually supports the scopes we're asking for.  The change as it stands could request scopes from the metadata server that it does not support, and bypass the check we do in _ScopesFromMetadataServer to ensure they're supported and output a useful error message.

craigcitro: ok, on another reading, this was indeed overzealous.

thobrla:
LGTM

PR #44, opened 2015-08-13
Use the Application Default Credential if available.

craigcitro:
Previously, we'd skip checking for the "well-known file" variant of the ADC if
present; now, we check for this before falling back to our file cache. In
particular, this involves adding a heuristic for when the set of scopes used
by gcloud are sufficient for our needs.

PTAL @silviulica

### DIFF
--- a/apitools/base/py/credentials_lib.py
+++ b/apitools/base/py/credentials_lib.py
@@ -54,6 +54,7 @@ def GetCredentials(package_name, scopes, client_id, client_secret, user_agent,
                    credentials_filename=None,
                    service_account_name=None, service_account_keyfile=None,
                    service_account_json_keyfile=None,
+                   skip_application_default_credentials=False,
                    api_key=None,  # pylint: disable=unused-argument
                    client=None,  # pylint: disable=unused-argument
                    oauth2client_args=None):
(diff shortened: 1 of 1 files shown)


craigcitro:
Also, @thobrla -- is there any chance this might be bad in any corner case for gsutil? or do you always bring your own credentials as far as apitools is concerned?

thobrla:
gsutil instantiates GceAssertionCredentials directly and never brings its own credentials.

That said, were we ever to use this code, its lack of thread safety could pose a problem.  The concern is multiple unprotected calls to _implicit_credentials_from_file.  Given that cache_file_lock already exists, consider guarding the call with it.

craigcitro:
good call @thobrla. done.

apitools/base/py/credentials_lib.py
```diff
@@ -129,6 +134,28 @@ def ServiceAccountCredentials(service_account_name, private_key, scopes,
...
+            # credentials, we don't want to pay for checking again.
+            credentials = gc._implicit_credentials_from_files()
+        except oauth2client.client.ApplicationDefaultCredentialsError:
+            return None
+    # If we got back a non-service-account credential, we need to use
+    # a heuristic to decide whether or not the application default
+    # credential will work for us. We assume that if we're requesting
+    # cloud-platform, our scopes are a subset of cloud scopes, and the
+    # ADC will work.
+    if not isinstance(credentials, gc):
```
thobrla: Nit: this could be a single if check, since you return credentials if either is true.

craigcitro: yep -- but there was no way to format it so that the linter didn't complain. :wink: 

fixing, because i don't let no linter boss ME around.

craigcitro: fixed.

thobrla:
LGTM

PR #46, opened 2015-08-16
Add a system for registering credentials fetching methods.

craigcitro:
This allows us to add new ways to fetch credentials that apitools itself need
not know about.

PTAL @thobrla

### DIFF
--- a/apitools/base/py/credentials_lib.py
+++ b/apitools/base/py/credentials_lib.py
@@ -48,64 +48,55 @@ def SetCredentialsCacheFileLock(lock):
     cache_file_lock = lock
 
 
-# TODO(craigcitro): Expose the extra args here somewhere higher up,
-# possibly as flags in the generated CLI.
+# List of additional methods we use when attempting to construct
+# credentials. Users can register their own methods here, which we try
+# before the defaults.
+_CREDENTIALS_METHODS = []
+
+
+def _RegisterCredentialsMethod(method, position=None):
+    """Register a new method for fetching credentials.
+
+    This new method should be a function with signature:
+      client_info, **kwds -> Credentials or None
+    This method can be used as a decorator, unless position needs to
+    be supplied.
+
+    Note that method must *always* accept arbitrary keyword arguments.
+
+    Args:
+      method: New credential-fetching method.
+      position: (default: None) Where in the list of methods to
+        add this; if None, we append. In all but rare cases,
+        this should be either 0 or None.
+    Returns:
+      method, for use as a decorator.
+
+    """
+    if position is None:
+        position = len(_CREDENTIALS_METHODS)
+    else:
+        position = min(position, len(_CREDENTIALS_METHODS))
+    _CREDENTIALS_METHODS.insert(position, method)
+    return method
+
+
 def GetCredentials(package_name, scopes, client_id, client_secret, user_agent,
                    credentials_filename=None,
-                   service_account_name=None, service_account_keyfile=None,
-                   service_account_json_keyfile=None,
-                   skip_application_default_credentials=False,
                    api_key=None,  # pylint: disable=unused-argument
                    client=None,  # pylint: disable=unused-argument
-                   oauth2client_args=None):
+                   oauth2client_args=None,
+                   **kwds):
     """Attempt to get credentials, using an oauth dance as the last resort."""
     scopes = util.NormalizeScopes(scopes)
-    if ((service_account_name and not service_account_keyfile) or
-            (service_account_keyfile and not service_account_name)):
-        raise exceptions.CredentialsError(
-            'Service account name or keyfile provided without the other')
-    # TODO(craigcitro): Error checking.
     client_info = {
         'client_id': client_id,
         'client_secret': client_secret,
-        'scope': ' '.join(sorted(util.NormalizeScopes(scopes))),
+        'scope': ' '.join(sorted(scopes)),
         'user_agent': user_agent or '%s-generated/0.1' % package_name,
     }
-    service_account_kwargs = {
-        'user_agent': client_info['user_agent'],
-    }
-    if service_account_json_keyfile:
-        with open(service_account_json_keyfile) as keyfile:
-            service_account_info = json.load(keyfile)
-        account_type = service_account_info.get('type')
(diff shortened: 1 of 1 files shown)


apitools/base/py/credentials_lib.py
```diff
@@ -48,65 +48,57 @@ def SetCredentialsCacheFileLock(lock):
...
+    This new method should be a function with signature:
+      client_info, **kwds -> Credentials or None
+    This method can be used as a decorator, unless position needs to
+    be supplied.
+
+    Note that method must *always* accept arbitrary keyword arguments.
+
+    Args:
+      method: New credential-fetching method.
+      position: (default: None) Where in the list of methods to
```
thobrla: Is there a working case where append / prepend aren't sufficient?  Inserting into a specific position seems ripe for compatibility problems when a dependency alters its number of credentials (or worse, inserts positionally itself).

apitools/base/py/credentials_lib.py
```diff
@@ -48,65 +48,57 @@ def SetCredentialsCacheFileLock(lock):
...
-        return credentials
-    credentials = GceAssertionCredentials.Get(scopes)
-    if credentials is not None:
-        return credentials
-    if not skip_application_default_credentials:
-        credentials = _GetApplicationDefaultCredentials(scopes)
+    for method in _CREDENTIALS_METHODS:
+        credentials = method(client_info, **kwds)
         if credentials is not None:
+            print('RETURNING CREDENTIALS: ', credentials)
```
thobrla: Move this print to a debug-only output.

apitools/base/py/credentials_lib.py
```diff
@@ -48,65 +48,57 @@ def SetCredentialsCacheFileLock(lock):
...
-    if ((service_account_name and not service_account_keyfile) or
-            (service_account_keyfile and not service_account_name)):
-        raise exceptions.CredentialsError(
-            'Service account name or keyfile provided without the other')
-    # TODO(craigcitro): Error checking.
     client_info = {
         'client_id': client_id,
         'client_secret': client_secret,
-        'scope': ' '.join(sorted(util.NormalizeScopes(scopes))),
+        'scope': ' '.join(sorted(scopes)),
```
thobrla: Any particular reason to not dedupe here?

apitools/base/py/credentials_lib.py
```diff
@@ -48,65 +48,57 @@ def SetCredentialsCacheFileLock(lock):
     cache_file_lock = lock
 
 
-# TODO(craigcitro): Expose the extra args here somewhere higher up,
-# possibly as flags in the generated CLI.
+# List of additional methods we use when attempting to construct
+# credentials. Users can register their own methods here, which we try
+# before the defaults.
+_CREDENTIALS_METHODS = []
```
thobrla: A similar approach (allowing arbitrary auth plugins) is used in boto.  One danger is that it makes backwards compatibility testing of top-level methods very difficult, as it may impact user-authored credential methods.

This doesn't block the PR, but it's something to be aware of.  If there are any compatibility guarantees (or disclaimers) that you want to call you, this would be a good place.

apitools/base/py/credentials_lib.py
craigcitro: i have two internal use cases for this -- i want it to be *after* service account, but *before* others.

apitools/base/py/credentials_lib.py
craigcitro: ah, this was debugging noise. dropped.

apitools/base/py/credentials_lib.py
craigcitro: do you mean not call `util.NormalizeScopes` again? mostly because we did 3 lines above.

(or are you pointing out something else and i'm missing it?)

apitools/base/py/credentials_lib.py
craigcitro: that's interesting to hear -- i didn't realize boto also did things this way, but it makes sense.

my primary need is to invert the dependencies in some internal code; i could actually make these methods internal-only for now (they're not listed in `__all__` yet, on purpose), if you think i'm going to regret it ...

thobrla: As long as you're aware of what you're getting into, I think a comment is fine.

apitools/base/py/credentials_lib.py
thobrla: No, I just missed the call above; sorry.

apitools/base/py/credentials_lib.py
thobrla: If the caller needs to control it to that extent, what do you think about returning a list of methods to the caller and requiring them to rebuild with their own order, possibly disincluding certain methods? Something along the lines of this, from the caller's perspective:

default_methods = credentials_lib.GetCredentialsMethods()
get_creds_methods = []
get_creds_methods.append(default_methods['GetServiceAccountCredentials'])
get_creds_methods.append(_UserSpecifiedCredentialsMethod)
for method_name, method in default_methods.iteritems():
  if (method_name != 'GetServiceAccountCredentials' and
      method_name != 'GetApplicationDefaultCredentials'):
    get_creds_methods.append(method)

credentials_lib.SetCredentialsMethods(get_creds_methods)

I don't think we can make it easy on the caller when apitools adds a new method, but at the least we could make the ordering explicit and centralized.

craigcitro: so, two thoughts:

1. given that i don't forsee any other methods being added soon, i'm interested in keeping it simple.

2. in fact, i'm OK saying that how we add methods is an implementation detail, and waiting until i have a good reason to make it public.

so i'm making the methods private for now, and then i feel safer with the admittedly strange interface. if and when someone else presents a good case for sharing them, we'll see what we need.

apitools/base/py/credentials_lib.py
craigcitro: i made them module-private for now -- this being python, users **can** still do strange things. i just won't feel responsible for it. :wink:

apitools/base/py/credentials_lib.py
thobrla: Works for me.

thobrla:
LGTM

PR #47, opened 2015-08-21
Internal hooks

craigcitro:
PTAL @thobrla

this adds the last few hooks I need to be able to remove the version skew between the two internal copies of apitools.


thobrla:
LGTM

PR #48, opened 2015-08-26
Fix OverflowError in GetRange on 32-bit Python

thobrla:


craigcitro:
i almost wonder if we shouldn't remove the `__len__` attribute to avoid this kind of pain for ourselves. ;)

PR #50, opened 2015-09-04
Let gen_client run on windows

cherba:
Remove trivial dependence of gen_client on base_cli module, as the former depends on readline which is not supported on windows.


cherba:
I signed it!

PR #52, opened 2015-09-10
Ensure apitools.gen.util.Chdir cleans up.

craigcitro:
Previously, if a test failed, later tests using `Chdir` would fail, since
they're in a now-nonexistent directory, and `os.getcwd()` fails.

PTAL @cherba


cherba:
LGTM.

PR #53, opened 2015-09-10
Remove google.apputils/gflags usage in favor of argparse.

cherba:
This PR converts apitools.gen.gen/gen_client.py to use python standard argparse for command line argument parsing. Note that now type of generation argument must occur last on the command line.

### DIFF
--- a/apitools/gen/gen_client.py
+++ b/apitools/gen/gen_client.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 """Command-line interface to gen_client."""
 
+import argparse
 import contextlib
 import json
 import logging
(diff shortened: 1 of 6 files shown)


apitools/gen/gen_client.py
```diff
@@ -221,86 +142,172 @@ def _WriteSetupPy(codegen):
...
+    pip_package_parser = subparsers.add_parser(
+        'pip_package', help='Generate apitools client pip package')
+    pip_package_parser.set_defaults(func=GeneratePipPackage)
+
+    proto_parser = subparsers.add_parser(
+        'proto', help='Generate apitools client protos')
+    proto_parser.set_defaults(func=GenerateProto)
+
+    args = parser.parse_args(argv[1:])
+    args.func(args)
```
craigcitro: we probably want to get a return value here, and then return either that or 0 (since we might fall off the end of a function and return None in places).

then we want to `sys.exit(main())` below.

craigcitro:
wahoo! this is super awesome. just one small nit with return values and we should be good to go.

apitools/gen/gen_client.py
cherba: Done.

craigcitro:
w00t!

PR #54, opened 2015-09-11
Update our use of protorpc metaclasses.

craigcitro:
We need to subclass a given protorpc metaclass in extra_types.py; as luck
would have it, that had to be tweaked and updated for python3, which means
we've got to do something here, too.

This updates us to require a new minimum protorpc version, and switches from
`protorpc.messages.Field.__metaclass__` to `protorpc.messages._FieldMeta` for
the metaclass we inherit from.

PTAL @cherba


cherba:
LGTM.

PR #58, opened 2015-09-18
Protorpclite: forking the protorpc bits we need.

craigcitro:
PTAL @cherba @thobrla

This is a significant change: I've pulled the parts of protorpc that we use directly into apitools, and dropped the dependency on protorpc. This has a number of advantages, primarily:

1. it's one less dependency for us. this also means one less library to monitor, deal with permissions, etc.
2. there's quite a bit of code that works around particular protorpc choices -- we can't fix those in protorpc, since there are other users, but we can just fix them here.

Note that this will entail one medium-but-mechanical work item: `gcloud` tests that include a `TimeZoneOffset` will need a tweak for the module name.


thobrla:
Implementation and general approach LGTM.

cherba:
Thanks, LGTM.

PR #59, opened 2015-09-21
Fix an issue with client generation, and a test.

craigcitro:
Previously, the test of client generation was failing, as the exit code for
`./<generated_cli> help` was indistinguishable from a real error. This only
showed up as output running tests.

First, we fix the test by checking the output for a traceback. Not foolproof,
but a good start.

Next, we fix the code itself, by no longer letting `--outdir` be used as the
value for `--root_package_dir`. We have both flags; it's a bit silly to have
wacky logic that tries to use the value of one to outsmart the caller.
Instead, we're explicit, and if we have to repeat ourselves, no big deal. This
also allows us to delete `apitools.gen.util.GetPackage`.

PTAL @cherba


cherba:
LGTM.

PR #60, opened 2015-09-29
Improve error reporting in BaseApiClient.DeserializeMessage

b-daniels:
Convert ValueError to InvalidDataFromServerError in BaseApiClient.DeserializeMessage

Currently, the ValueErrors raised by json.loads are not handled by DeserializeMessage, meaning that we're losing the text of the source message in error logs when the JSON is malformed.


craigcitro:
Thanks!

PR #62, opened 2015-10-16
Add global_params pass-through support to list_pager.

calbach:
This enables additional request parameterization of the underlying request, e.g. partial response filtering.

### DIFF
--- a/apitools/base/py/list_pager.py
+++ b/apitools/base/py/list_pager.py
@@ -9,7 +9,7 @@ __all__ = [
 
 
 def YieldFromList(
-        service, request, limit=None, batch_size=100,
+        service, request, global_params=None, limit=None, batch_size=100,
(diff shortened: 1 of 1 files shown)


apitools/base/py/list_pager.py
```diff
@@ -46,7 +48,7 @@ def YieldFromList(
     setattr(request, current_token_attribute, None)
     while limit is None or limit:
         response = getattr(service, method)(request)
```
craigcitro: actually, shouldn't the argument get threaded in here?

calbach: Oops! Thanks. I did a pretty bad copy by-eye of the working code. I guess we'll see if travis would have caught that one.

Fixed.

craigcitro:
awesome -- thanks!

PR #64, opened 2015-10-23
Tolerate remappings if the mapped value is the same.

vilasj:


craigcitro:
(googlebot wants you to join the google org)

craigcitro:
can you add a quick test?

vilasj:
Added a test. Craig, let me know if there is something appropriate that I can assert at the end of the test.

craigcitro:
that looks good -- you could always have an assertion to confirm it still fails on a new remapping?

craigcitro:
LGTM in any event, you can merge either before or after another test as you prefer :wink:

PR #66, opened 2015-10-27
actually call testValidateElement

gbin:
corrected also a couple typos in the test itself that prevented it to run.


craigcitro:
:grin: 

thanks!

PR #67, opened 2015-10-27
Relax the int vs float constraint

gbin:
As in JSON encoding they are equivalent, we can let them through.

### DIFF
--- a/apitools/base/protorpclite/messages.py
+++ b/apitools/base/protorpclite/messages.py
@@ -1321,7 +1321,9 @@ class Field(six.with_metaclass(_FieldMeta, object)):
         Raises:
           ValidationError if value is not expected type.
         """
-        if not isinstance(value, self.type):
+        if not (isinstance(value, self.type) or
+            (isinstance(value, (float, int)) and
+             self.type in (float, int))):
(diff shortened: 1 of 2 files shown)


apitools/base/protorpclite/messages.py
```diff
@@ -1321,7 +1321,9 @@ def validate_element(self, value):
         Raises:
           ValidationError if value is not expected type.
         """
-        if not isinstance(value, self.type):
+        if not (isinstance(value, self.type) or
+            (isinstance(value, (float, int)) and
```
craigcitro: this test is looking a little hard to read -- can we just have

    if isinstance(value, int) and self.type == float:
      return

as the first few lines inside the `if`?

(note that i'm also trimming the logic to just what we need for now -- float -> int conversion is more worrisome, due to precision/rounding questions, i think?)

gbin:
PTAL

craigcitro:
LGTM, but the linter wants you to fix a long line (in messages_test). tweak that and i'll merge.

thanks for cleaning this up!

gbin:
PTAL.

PR #69, opened 2015-10-28
Make `tox -e lint` faster in the common case.

craigcitro:
This cuts lint checks down from ~28s to ~4s on my machine (with a massive
sample size of N=3).

PTAL @gbin -- this should at least *mostly* improve the situation you were hitting. :wink:

### DIFF
--- a/run_pylint.py
+++ b/run_pylint.py
@@ -96,7 +96,7 @@ def is_production_filename(filename):
                 filename.startswith('regression'))
 
 
-def get_files_for_linting(allow_limited=True):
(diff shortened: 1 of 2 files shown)


gbin:
better for me but still...

tox -e lint  23.38s user 0.99s system 99% cpu 24.374 total

weirdly enough it goes through pep8 quite fast then stay there for ~20s before generating "____... summary" then it congratules me

craigcitro:
so the problem here is that it's calling gpylint on every file (and i think collating results? maybe going out for coffee?)

there's more to do, but i'll take at least a small improvement for now ...

PR #70, opened 2015-11-03
Support custom check_response_func in BaseApiClient

jlowdermilk:
This lets apis overwrite the default check_response_func used by MakeRequest, e.g. to change retry behavior for specific http error codes.

### DIFF
--- a/apitools/base/py/base_api.py
+++ b/apitools/base/py/base_api.py
@@ -237,7 +237,8 @@ class BaseApiClient(object):
     def __init__(self, url, credentials=None, get_credentials=True, http=None,
                  model=None, log_request=False, log_response=False,
                  num_retries=5, max_retry_wait=60, credentials_args=None,
-                 default_global_params=None, additional_http_headers=None):
+                 default_global_params=None, additional_http_headers=None,
+                 check_response_func=None):
         _RequireClassAttrs(self, ('_package', '_scopes', 'messages_module'))
         if default_global_params is not None:
             util.Typecheck(default_global_params, self.params_type)
@@ -263,6 +264,7 @@ class BaseApiClient(object):
         self.__include_fields = None
 
         self.additional_http_headers = additional_http_headers or {}
+        self.__check_response_func = check_response_func
 
         # TODO(craigcitro): Finish deprecating these fields.
         _ = model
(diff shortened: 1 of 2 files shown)


jlowdermilk:
Finally appeased the linter. cc @craigcitro @thobrla

craigcitro:
also pinging @cherba and @vilasj if either of them has time to take a look

thobrla:
LGTM

apitools/base/py/base_api.py
```diff
@@ -376,6 +378,15 @@ def num_retries(self, value):
         self.__num_retries = value
 
     @property
+    def check_response_func(self):
+        return self.__check_response_func
+
+    @check_response_func.setter
+    def check_response_func(self, value):
+        if value:
```
vilasj: Why is this check required?

apitools/base/py/base_api.py
```diff
@@ -376,6 +378,15 @@ def num_retries(self, value):
         self.__num_retries = value
 
     @property
+    def check_response_func(self):
```
vilasj: Looks like we don't really need this property? It is only being used privately except in the test? In the test it can be passed through the constructor anyway?

apitools/base/py/base_api.py
```diff
@@ -685,9 +696,14 @@ def _RunMethod(self, method_config, request, global_params=None,
             http = self.__client.http
             if upload and upload.bytes_http:
                 http = upload.bytes_http
+            opts = {
```
vilasj: Why create opts and expand it instead of just passing through the three params?

apitools/base/py/base_api.py
jlowdermilk: It's not, removed.

apitools/base/py/base_api.py
jlowdermilk: I figured it would be easier to use if you can set it set after instantiating a generated api. Otherwise generated api has to plumb it through in its constructor which may be annoying.

thobrla: Setter seems useful because you can modify the client temporarily for a single call or a class of calls.

apitools/base/py/base_api.py
jlowdermilk: To avoid putting the MakeRequest call in both branches of an if/else, since we don't want to explicitly pass `check_response_func=None`.

apitools/base/py/base_api.py
vilasj: sg

apitools/base/py/base_api.py
vilasj: ah ok sg

vilasj:
LGTM

jlowdermilk:
Squashed.

jlowdermilk:
@craigcitro can haz merge?

craigcitro:
sure! i thought @vilasj or @thobrla woudl merge. ;)

PR #77, opened 2015-11-23
Generate client libs with apitools pinned to the version used to generate them if v >= 0.5

silviulica:
This change does three things:
1) Client packages will have the version APITOOLS_VERSION.CLIENT_REVISION_DATE
   (e.g., 0.4.21.20151102)
2) Client packages have the dependency on google-apitools pinned to the
   version used to generate them.
3) Bumped up version of google-apitools to 0.4.21 to cover the few client
   packages out there that have an older format for version (0.4.REVISION_DATA).
   Version 0.4.21 will be newer than versions like 0.4.2015MMDD so more recent
   packages will be installed.


craigcitro:
this LGTM -- just have to sort out the merge conflicts

silviulica:
I am going to update this because there are corner cases where this will not work.
It is better to leave 0.4.x behavior as it is and do this new stuff only for >= 0.5.x. Will explain the new pull request.

silviulica:
PTAL Travis/Craig
I had to add some tests to get a bit of test coverage on the changes.

apitools/gen/gen_client_lib.py
```diff
@@ -223,8 +239,10 @@ def WriteSetupPy(self, out):
         printer('import setuptools')
         printer('REQUIREMENTS = [')
         with printer.Indent(indent='    '):
-            # TODO(craigcitro): Have this track apitools' version.
-            printer('"google-apitools>=0.4.8",')
+            if self.apitools_version.startswith('0.4.'):
+                printer('"google-apitools>=0.4.8,<0.5",')
```
craigcitro: do we need to worry about this case? we won't be regenerating old clients with the new code, right?

craigcitro:
LGTM modulo one little question

PR #80, opened 2015-12-07
Import on demand credentials_lib in base_api.

cherba:
credentials_lib depends on oauth2client which had support for webflow authentication, which brings a lot of dependencies, all not used if user of the client manages its own credentials.


craigcitro:
LGTM (feel free to merge once travis passes)

cherba:
Strange, travis seems to be stuck. In my fork travis failed on existing lint errors.

craigcitro:
oh, right, I think Vilas hit the same issue with lint.

PR #82, opened 2015-12-08
In client mock explicitly import its dependencies.

cherba:
This helps readability, but also reduces imports which otherwise might not be needed.


craigcitro:
LGTM

PR #84, opened 2016-01-11
Update encoding.py with some bug fixes

bradfriedman:
Normalize the ordering of the return value of encode_message in encoding.py. Also, some miscellaneous cleaning up.

### DIFF
--- a/apitools/base/py/encoding.py
+++ b/apitools/base/py/encoding.py
@@ -157,7 +157,7 @@ def MessageToRepr(msg, multiline=False, **kwargs):
 
     """
 
-    # TODO(user): craigcitro suggests a pretty-printer from apitools/gen.
+    # TODO(jasmuth): craigcitro suggests a pretty-printer from apitools/gen.
 
     indent = kwargs.get('indent', 0)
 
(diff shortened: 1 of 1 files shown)


apitools/base/py/encoding.py
```diff
@@ -406,7 +406,7 @@ def _DecodeUnknownMessages(message, encoded_message, pair_type):
         new_values.append(new_pair)
     return new_values
 
-
+a
```
craigcitro: stray character

craigcitro:
Two things:

1. Maybe update the commit description to say what you're fixing -- IIUC, this is just normalizing the order on the return value, right?

2. I'll send a separate email with info on how to make cla-signer happy. :wink:

bradfriedman:
I feel like the lint error here is not useful in this context -- it seems idiomatic to me to check whether a type is in a list of types. Thoughts? Should I disable that rule with an in-line comment?

************* Module apitools.base.py.encoding
C:331,11: Using type() instead of isinstance() for a typecheck. (unidiomatic-typecheck)

craigcitro:
yeah, i agree that the linter warning is spurious here. (we really need the type to be a key in our dict.) you can disable or just bind to a temp var (which confuses the linter :wink:).

i'd say do that and squash, and i'll force cla-signer to wake up if it doesn't chime in.

apitools/base/py/encoding.py
```diff
@@ -328,12 +328,15 @@ def encode_message(self, message):
         if isinstance(message, messages.FieldList):
             return '[%s]' % (', '.join(self.encode_message(x)
                                        for x in message))
-        message_type = type(message)
-        if message_type in _CUSTOM_MESSAGE_CODECS:
+
+        # pylint: disable=unidiomatic-typecheck
```
craigcitro: small nit: this actually disables the warning for the rest of this block; if you put it at the end of the line, it'll only disable it on that line.

(happily, the linter should be OK with a long line that's due to a pylint disable.)

bradfriedman: I actually did that at first, then the linter complained because the line was too long :/

PR #88, opened 2016-02-15
Drop gflags/appcommands in oauth2l, support py3.

craigcitro:
We'd only used gflags in oauth2l for vestigial reasons (read: I first wrote
the code internally), but someone ran into related install trouble this
weekend. Swapping out argparse was easy, and as a nice by-product, we pick up
python3 support in oauth2l.

(As a bonus, there's slightly more documentation for a net drop in LOC. w00t.)

PTAL @cherba @wora

### DIFF
--- a/apitools/scripts/oauth2l.py
+++ b/apitools/scripts/oauth2l.py
@@ -36,8 +36,9 @@ some sample use:
 
 The `header` command is designed to be easy to use with `curl`:
 
-    $ curl "$(oauth2l header bigquery)" \
-           'https://www.googleapis.com/bigquery/v2/projects'
+    $ curl -H "$(oauth2l header bigquery)" \\
+      'https://www.googleapis.com/bigquery/v2/projects'
+    ... lists all projects ...
 
 The token can also be printed in other formats, for easy chaining
 into other programs:
@@ -49,7 +50,9 @@ into other programs:
 
 """
 
-import httplib
+from __future__ import print_function
+
+import argparse
 import json
 import logging
 import os
@@ -57,14 +60,11 @@ import pkgutil
 import sys
 import textwrap
 
-import gflags as flags
-from google.apputils import appcommands
 import oauth2client.client
+from six.moves import http_client
 
 import apitools.base.py as apitools_base
-from apitools.base.py import cli as apitools_cli
 
-FLAGS = flags.FLAGS
 # We could use a generated client here, but it's used for precisely
 # one URL, with one parameter and no worries about URL encoding. Let's
 # go with simple.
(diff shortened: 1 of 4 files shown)


craigcitro:
Ah, coveralls is complaining because we no longer have tests covering app2/base_cli; I think this is okay, since I'd like to ultimately drop those (or convert them to be gflags-free).

apitools/scripts/oauth2l.py
```diff
@@ -98,12 +85,13 @@ def GetDefaultClientInfo():
     }
 
 
-def GetClientInfoFromFlags():
-    """Fetch client info from FLAGS."""
-    if FLAGS.client_secrets:
-        client_secrets_path = os.path.expanduser(FLAGS.client_secrets)
+def GetClientInfoFromFlags(args):
```
cherba: Seems like only one value is needed, can you pass that instead?

apitools/scripts/oauth2l.py
```diff
@@ -173,20 +167,20 @@ def _ValidateToken(access_token):
     return bool(_GetTokenScopes(access_token))
 
 
-def FetchCredentials(scopes, client_info=None, credentials_filename=None):
+def FetchCredentials(args, client_info=None, credentials_filename=None):
```
cherba: can you pass in values instead of args?

apitools/scripts/oauth2l.py
```diff
@@ -194,131 +188,152 @@ def FetchCredentials(scopes, client_info=None, credentials_filename=None):
...
-class _Header(apitools_cli.NewCmd):
-
-    """Print credentials for a header."""
-
-    usage = 'header <scope> [<scope> ...]'
-
-    def RunWithArgs(self, *scopes):
-        """Fetch a valid access token and display it formatted for a header."""
-        print _Format('header', FetchCredentials(scopes))
+def Email(args):
```
cherba: The name of the function is not indicative on what it does.
Also can it be private?

apitools/scripts/oauth2l.py
craigcitro: done.

apitools/scripts/oauth2l.py
craigcitro: made them all private -- they're just paired with the names of the corresponding subcommands.

apitools/scripts/oauth2l.py
craigcitro: done.

apitools/scripts/oauth2l.py
craigcitro: actually, not done -- but made the function private instead. (nothing else depends on this file, so i'm not too worried about wanting to change the API later if it needs to.)

craigcitro:
PTAL

craigcitro:
Ah, it looks like there's a new oauth2client release that moved some module names around. (Travis is going to fail tests.)

craigcitro:
PTAL -- tests are passing again (after the oauth2client fix)

cherba:
LGTM

wora:
LGTM

PR #89, opened 2016-02-17
Update oauth2client imports for the new release.

craigcitro:
We could pin, but this seems more useful, especially given the small surface
we use in oauth2client.

(Also added a handful of lint disables, since the order it expected was for
the try/except imports to come in the middle of the list of imports.)

PTAL @cherba @thobrla

### DIFF
--- a/apitools/base/py/credentials_lib.py
+++ b/apitools/base/py/credentials_lib.py
@@ -25,9 +25,6 @@ import threading
 import httplib2
 import oauth2client
 import oauth2client.client
-import oauth2client.gce
-import oauth2client.locked_file
-import oauth2client.multistore_file
 import oauth2client.service_account
 from oauth2client import tools  # for gflags declarations
 from six.moves import http_client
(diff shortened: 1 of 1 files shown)


apitools/base/py/credentials_lib.py
```diff
@@ -36,8 +33,13 @@
 from apitools.base.py import exceptions
 from apitools.base.py import util
 
+# pylint: disable=wrong-import-order,ungrouped-imports
+try:
+    from oauth2client import gce, locked_file, multistore_file
```
thobrla: Add a comment about the versions this spans and at what point it's safe to remove?

thobrla:
Approach seems reasonable.

apitools/base/py/credentials_lib.py
craigcitro: Done.

craigcitro:
Thanks!

craigcitro:
(This already passed tests on my repo, so merging.)

thobrla:
Post-hoc LGTM

PR #90, opened 2016-02-17
Deprecate X-Google-Metadata-Request in favor new Metadata-Flavor header

Galabar001:
We are deprecating the use of the X-Google-Metadata-Request header in favor
of the latest header, Metadata-Flavor.

The transition is explained here:

https://cloud.google.com/compute/docs/metadata#transitioning


craigcitro:
I think we actually want to close this another way (see #91).

Galabar001:
Sounds good.

Thanks!

PR #91, opened 2016-02-17
Cleanup some dead test code.

craigcitro:
It turns out an ancient refactor (merging the gsutil fork) turned some tests
into dead code; I noticed that something was awry when the GCE header name was
updated but tests didn't fail.

The fix was easy (just drop the dead code and add something new), and I bumped
up coverage a wee bit in the process. (It'd be easy to keep going, but I hate
mocks, and think most of these tests have little value.)

Closes #90 (via different means).

PTAL @thobrla

### DIFF
--- a/apitools/base/py/credentials_lib_test.py
+++ b/apitools/base/py/credentials_lib_test.py
@@ -13,29 +13,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import re
-
 import mock
 import six
-from six.moves import http_client
 import unittest2
 
 from apitools.base.py import credentials_lib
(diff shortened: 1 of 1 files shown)


thobrla:
LGTM

Galabar001:
I think the ComputeEngineCredentials.java and MockMetadataServerTransport.java files are out-of-sync internally.

Where should changes be made (in google3/ or on github)?

craigcitro:
@Galabar001 wrong repo? :wink:

Galabar001:
Garr!!! :)  So many things to change, so many incorrect repos to spam. :(

PR #92, opened 2016-02-19
Bump version to 0.5.0 and update apitools requirements to match that.

silviulica:
HI Craig,
Hopefully this will finally work. The client libs now will have a dependency on apitools that looks like this:
apitools>=major.minor.rest,~=major.minor
Let me know what you think.
Thanks!
Silviu


apitools/gen/gen_client_lib.py
```diff
@@ -240,10 +241,11 @@ def WriteSetupPy(self, out):
         printer('import setuptools')
         printer('REQUIREMENTS = [')
         with printer.Indent(indent='    '):
-            if self.apitools_version.startswith('0.4.'):
-                printer('"google-apitools>=0.4.8,<0.5",')
-            else:
-                printer('"google-apitools==%s",', self.apitools_version)
+            match = re.search(
```
craigcitro: the re works here, but i'd rather do something like

  parts = self.apitools_version.split('.')
  major = parts.pop(0)
  minor = parts.pop(0)

and use `major` and `minor` below. it's a little simpler, but it'll work even if we release a version 0.6 or 1.0 one day (which would fail to match the re).

setup.py
```diff
@@ -55,7 +55,7 @@
 if py_version < '2.7':
     REQUIRED_PACKAGES.append('argparse>=1.2.1')
 
-_APITOOLS_VERSION = '0.4.15'
+_APITOOLS_VERSION = '0.5.0'
```
craigcitro: can we set this to either 0.4.99 or 0.5.0a1 for now? i still want to see what happens with gsutil testing against the new version (in particular with the new oauth2client).

craigcitro:
LGTM, two little nits

silviulica:
Added the changes you requested to the pull request.  Please LGTM again just to be sure :-)
If you are ok I can do the merge , tag version and release.

craigcitro:
LGTM!

PR #93, opened 2016-02-22
Update service_account support for oauth2clientv2.

craigcitro:
This makes three changes:

* Make the code for loading either .json or .p12 credentials work under either
  oauth2client 2.0.0post1+ or oauth2client < 2.0.0. In particular, this mostly
  amounts to picking up a few renames from the oauth2client changelog.

* Drop `credentials_lib.ServiceAccount()`. From some searching, no one at all
  is using this function, and supporting it makes the case of .p12 keys
  significantly more awkward.

* Change `credentials_lib.ServiceAccountFromFile()` to handle .json keys, not
  .p12 keys. I'm fairly certain that apitools itself was the only caller of
  this code.

Tested by hand (via oauth2l).

PTAL @silviulica -- this should mean we're good to go on a release.

FYI @thobrla -- I've confirmed from looking at code that this won't break you, but I thought you might be interested in using some of the same code in gsutil if/when you want to support a newer oauth2client.

### DIFF
--- a/apitools/base/py/credentials_lib.py
+++ b/apitools/base/py/credentials_lib.py
@@ -25,7 +25,7 @@ import threading
 import httplib2
 import oauth2client
 import oauth2client.client
-import oauth2client.service_account
+from oauth2client import service_account
 from oauth2client import tools  # for gflags declarations
 from six.moves import http_client
 from six.moves import urllib
@@ -56,7 +56,6 @@ __all__ = [
     'GceAssertionCredentials',
     'GetCredentials',
     'GetUserinfo',
-    'ServiceAccountCredentials',
     'ServiceAccountCredentialsFromFile',
 ]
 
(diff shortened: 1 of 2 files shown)


thobrla:
LGTM, thanks for the heads-up.  We'll probably use this approach in gsutil (eventually).

silviulica:
LGTM 
Will this work with latest oauth2client? I saw some errors related to missing SignedJwtCredentials. Is this something that will be fixed with latest release of oauth2client or it is something we need to handle from now on?

silviulica:
One more question. Ready to release as 0.5.0 or still a pre-release?

craigcitro:
Yep, this works with the latest version of oauth2client. The two new blocks check for a class that should only be present in a sufficiently new version and either uses old or new code depending on what it finds.

That said, it occurs to me that I can just check the version. Tweak coming up.

Once this is in, we're good to release.

PR #95, opened 2016-03-08
Attempt to import newer versions of oauth2client libraries first.

Galabar001:
We have instances of this library being used in environments where only a
portion of the oauth2client library has been updated.  This change allows
us to find the latest versions of these libraries, when available.

### DIFF
--- a/apitools/base/py/credentials_lib.py
+++ b/apitools/base/py/credentials_lib.py
@@ -39,9 +39,19 @@ from apitools.base.py import util
 #
 # pylint: disable=wrong-import-order,ungrouped-imports
 try:
-    from oauth2client import gce, locked_file, multistore_file
+    from oauth2client.contrib import gce
 except ImportError:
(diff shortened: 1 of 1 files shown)


apitools/base/py/credentials_lib.py
```diff
@@ -39,9 +39,19 @@
 #
 # pylint: disable=wrong-import-order,ungrouped-imports
 try:
-    from oauth2client import gce, locked_file, multistore_file
+    from oauth2client.contrib import gce
```
cherba: Can gce, locked_file and multistore_file come from old and new locations at same time?

Galabar001: gce is separate.  We can coordinate the locked_file and multistore_file when we try to bring those in internally (that should be the only place where we see this issue -- github oauth2client is up-to-date).

craigcitro:
LGTM

PR #96, opened 2016-03-11
Cleanup package data, and fix a warning.

craigcitro:
This does two small-but-independent cleanups:

1. Fixes MANIFEST.in to always pick up the client_secrets.json file.

2. Tweaks a call in credentials_lib to avoid a (useful) warning in
oauth2client.contrib.gce.

PTAL @silviulica @cherba


thobrla:
LGTM

PR #105, opened 2016-05-09
Fix an issue with Python3 multipart encodings.

craigcitro:
As discovered in
https://github.com/GoogleCloudPlatform/gcloud-python/issues/1760, we were
mangling bytes when encoding them as part of a multipart upload request. The
fix is to switch from using `six.StringIO` to `six.BytesIO` in `transfer.py`.
The patch here is closely based on
https://github.com/GoogleCloudPlatform/gcloud-python/pull/1779.

PTAL @thobrla 

cc @dhermes -- Danny, if you have two minutes to take a look at the test, I would love a set of more-python3-savvy eyes. It fails before the patch and passes after, so it's definitely testing the right things, but I would be willing to believe I've managed to fool myself. 😉

### DIFF
--- a/apitools/base/py/transfer.py
+++ b/apitools/base/py/transfer.py
@@ -758,20 +758,24 @@ class Upload(_Transfer):
         # NOTE: We encode the body, but can't use
         #       `email.message.Message.as_string` because it prepends
         #       `> ` to `From ` lines.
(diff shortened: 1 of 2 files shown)


dhermes:
Took a look. No glaring issues.

thobrla:
LGTM

PR #107, opened 2016-05-18
Allow clients to set retry_func for MakeRequest

houglum:
This change:
- Adds an item to ExceptionRetryArgs containing the total number of
  seconds the user has waited since the first request attempt.
- Allows clients to specify a custom retry function for retry logic.

My use case for these changes involves using a wrapper function which performs
some additional logging if total_wait_sec is high enough, then calls the
default retry function.

### DIFF
--- a/apitools/base/py/base_api.py
+++ b/apitools/base/py/base_api.py
@@ -237,7 +237,7 @@ class BaseApiClient(object):
                  model=None, log_request=False, log_response=False,
                  num_retries=5, max_retry_wait=60, credentials_args=None,
                  default_global_params=None, additional_http_headers=None,
-                 check_response_func=None):
+                 check_response_func=None, retry_func=None):
         _RequireClassAttrs(self, ('_package', '_scopes', 'messages_module'))
         if default_global_params is not None:
             util.Typecheck(default_global_params, self.params_type)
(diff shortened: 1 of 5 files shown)


houglum:
For extra context, I plan to use this functionality in gsutil

apitools/base/py/base_api.py
```diff
@@ -398,6 +399,14 @@ def max_retry_wait(self, value):
                 'max_retry_wait must be a postiive integer')
         self.__max_retry_wait = value
 
+    @property
```
craigcitro: If there's no validation or anything here, you can just make this a public property.

craigcitro:
LGTM -- two things:

1. @thobrla can you PTAL for gsutil-ness and merge?
2. for the lint warning, feel free to add `too-many-arguments` to the list of ignores in the base `default.pylintrc` file.

thobrla:
Looks good, but let's get the lint warning disabled before we merge so that travis-ci is happy.

apitools/base/py/base_api.py
houglum: Sure. I was just trying to keep things uniform, as the check_response_func property also had a getter/setter.  I agree that they're unnecessary; making both of them public properties.

houglum:
Added to default.pylintrc and addressed the public property comment as well.

thobrla:
Squash commits and we should be good to merge (I assume you ran a smoke test with gsutil)?

houglum:
Squashed (and yes, I copied these changes over to my local gsutil repo and made sure it worked).

PR #110, opened 2016-05-24
Add generated client test sample with atomic names.

cherba:
This discovery doc uses atomic names, which is different from what we have supported so far. By checking in the discovery doc and its generated client, will allow in follow ups to make changes to generation so that names structure is exposed.


apitools/gen/gen_client_test.py
```diff
@@ -43,34 +43,43 @@ def testHelp_NotEnoughArguments(self):
...
-    def testGenClient_SimpleDoc(self):
+    def _CheckGeneratedFiles(self, api_name):
         with test_utils.TempDir() as tmp_dir_path:
             gen_client.main([
                 gen_client.__file__,
                 '--generate_cli',
                 '--init-file', 'empty',
-                '--infile', GetTestDataPath('dns', 'dns_v1.json'),
+                '--infile',
+                GetTestDataPath(api_name, api_name + '_v1.json'),
```
craigcitro: I'm overengineering a wee bit, but why not just pass `api_version` as another argument? 

I'm worried about the next person who adds a file, and says "wait, why is it looking for v1?"

craigcitro:
LGTM

apitools/gen/gen_client_test.py
cherba: Done.

PR #111, opened 2016-05-31
Consolidate gen test samples.

cherba:
This PR reorganizes generates test sample files.
  * Moves dns and iam generated clients to samples/ directory, as where storage client was generated.
  * Re-creates storage client from latest (now checked in) storage discovery doc.
  * Adds a script to regenerate all sample clients.
  * Adds a test to make sure all sample clients are up to date.


craigcitro:
LGTM

PR #112, opened 2016-06-03
Fix an ADC-related bug, and make an install tweak.

craigcitro:
When getting the application default credentials from a file, we need to make
sure we call create_scoped(), otherwise we'll get back invalid access tokens.

While I was here, I dropped the hook to add an oauth2l script, since it's now
its own package on PyPI. I'll drop the code in a separate PR.

PTAL @wora


wora:
LGTM

PR #114, opened 2016-06-03
Expose flat_path property in generated client services.

cherba:
flat_path method property exposes the structure of atomic names. This is helpful in automating construction of resource names from constituent parts.


craigcitro:
LGTM

PR #115, opened 2016-06-23
Move fusionteables sample to samples directory.

cherba:
This would allow these to get regenerated everytime generator is changed.


craigcitro:
LGTM

PR #116, opened 2016-06-24
Make method_config class level accessible.

cherba:
This PR tries to address a few goals simultaneously:
  1. Make method configs accessible statically, not requiring client initialization.
  1. Do not create all method configs upon import
  1. Do not create all method configs during client instantiation.

While method configs are relatively lightweight in grand schema of things, there can be quite a few of them. Some generated clients have 300-400 and sometimes applications need to use multiple such clients.
Since they are based on proto messages they are more expensive (measured upwards 20x) to instantiate as opposed to plain class or namedtuple.

For applications where startup time is important instantiating all these method_config instances during import time becomes prohibitive, running in over 100ms. At the same time there cases where structure of certain requests (like Get) needs to be introspected (for example when trying to parse a uri), without ever creating a client.

Longer term this extra functionality for managing resource ids can be (should be?) part of apitools.

This PR does not change existing public interface, and has been tested against heavy users of this library.

### DIFF
--- a/apitools/base/py/base_api.py
+++ b/apitools/base/py/base_api.py
@@ -472,7 +472,23 @@ class BaseApiService(object):
         return self.__client
 
     def GetMethodConfig(self, method):
-        return self._method_configs[method]
+        """Returns service cached method config for given method."""
+        method_config = self._method_configs.get(method)
+        if method_config:
+            return method_config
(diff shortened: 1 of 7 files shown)


craigcitro:
LGTM!

PR #125, opened 2016-08-08
Expose client mock url and http properties.

cherba:
Real apitools client has public url and http properties.
This ensures that these properties work in mocked client.


craigcitro:
lgtm

PR #128, opened 2016-09-29
Update YieldFromList to support APIs without a batch_size_attribute

benhead:
Cloud Bigtable services generally don't support a user-specified batch size in their List request objects. With this edit, YieldFromList can work with them.

### DIFF
--- a/apitools/base/py/list_pager.py
+++ b/apitools/base/py/list_pager.py
@@ -52,14 +52,15 @@ def YieldFromList(
           response message holding the page token for the next page.
       batch_size_attribute: str, The name of the attribute in a
           response message holding the maximum number of results to be
-          returned.
+          returned. None if caller-specified batch size is unsupported.
 
     Yields:
       protorpc.message.Message, The resources listed by the service.
 
     """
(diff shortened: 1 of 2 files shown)


apitools/base/py/list_pager.py
```diff
@@ -52,14 +52,15 @@ def YieldFromList(
...
-          returned.
+          returned. None if caller-specified batch size is unsupported.
 
     Yields:
       protorpc.message.Message, The resources listed by the service.
 
     """
     request = encoding.CopyProtoMessage(request)
-    setattr(request, batch_size_attribute, batch_size)
+    if batch_size_attribute:
```
cherba: Would it be helpful to add a small unit test in `apitools/base/py/list_pager_test.py` to test the scenario when attribute is not set.

benhead: Done. Note to avoid breaking existing code, unset still defaults to 'maxResults' as the field name. Callers must explicitly set it to None to indicate no field should be set (presumably because no such field exists).

PR #131, opened 2016-10-04
Do not raise exception in client mock if one already active.

cherba:
Developers get confused by this. While some test runners would show the stacktrace for both errors, only latter exception is scraped by some UIs masking the original uncaught exception.

This change makes sure to let original exception to propagate even if mock has failed.

### DIFF
--- a/apitools/base/py/testing/mock.py
+++ b/apitools/base/py/testing/mock.py
@@ -23,6 +23,7 @@ as it's all done within the context of a mock.
 """
 
 import difflib
+import sys
 
 import six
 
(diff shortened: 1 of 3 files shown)


apitools/base/py/testing/mock.py
```diff
@@ -345,15 +347,17 @@ def Unmock(self):
         del self._url
         del self._http
 
-        if self._request_responses:
```
vilasj: This check is deleted here but does not seem to be performed below....

apitools/base/py/testing/mock.py
```diff
@@ -345,15 +347,17 @@ def Unmock(self):
...
-
         self.__client_class.IncludeFields = self.__real_include_fields
         self.__real_include_fields = None
 
+        requests = [(rq_rs.key, rq_rs.request)
+                    for rq_rs in self._request_responses]
+        self._request_responses = []
+
+        if requests:
+            if not supress and sys.exc_info()[1] is None:
```
vilasj: Why do you need the 'and sys.exc_info()[1] is None' check here? Isn't suppress already set appropriately using that check?

apitools/base/py/testing/mock.py
```diff
@@ -331,12 +332,13 @@ class Patched(self.__class__, self.__client_class):
...
-        self.Unmock()
-        if value:
+        is_active_exception = value is not None
+        self.Unmock(supress=is_active_exception)
+        if is_active_exception:
             six.reraise(exc_type, value, traceback)
         return True
 
-    def Unmock(self):
+    def Unmock(self, supress=False):
```
vilasj: s/supress/suppress

cherba: Done.

PR #133, opened 2016-10-14
Pin oauth2client to avoid using version 4.0.0

charlesccychen:
The 4.0.0 version of oauth2client removes locked_file.


craigcitro:
LGTM 

/cc @cherba or @vilasj to merge

PR #135, opened 2016-11-02
Fix 'IncludeFields' method.

przemekwitek:
Fix 'IncludeFields' method so that it leaves siblings of the cleared fields
unaffected.


przemekwitek:
I signed it!

PR #136, opened 2016-11-17
Fix a nasty encoding corner case with JsonValue.

craigcitro:
The json encoding code in protorpc had a bit of custom logic to handle the
case of repeated vs. non-repeated fields with the same line of code. However,
in the case of a JsonValue (aka `any` type in a discovery doc), it's possible
for a non-repeated value to itself be a list. In this case, we'd always
deserialize the whole list, and then ... only use the last value.

The fix here is to actually simplify the code a bit, and split the logic based
on whether the field is repeated, *not* whether the value is itself a list.

This PR is partly for demonstration -- we need a test.

/cc @sb2nov

### DIFF
--- a/apitools/base/protorpclite/protojson.py
+++ b/apitools/base/protorpclite/protojson.py
@@ -277,22 +277,20 @@ class ProtoJson(object):
                         'No variant found for unrecognized field: %s', key)
                 continue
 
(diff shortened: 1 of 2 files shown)


sb2nov:
@cherba PTAL

PR #137, opened 2016-12-19
Add limit support for limiting size of batch request.

cherba:
Some apis impose limit on batch requests.
See for example:
  https://cloud.google.com/compute/docs/api/how-tos/batch

### DIFF
--- a/apitools/base/py/batch.py
+++ b/apitools/base/py/batch.py
@@ -28,6 +28,7 @@ import uuid
 import six
 from six.moves import http_client
 from six.moves import urllib_parse
+from six.moves import range  # pylint: disable=redefined-builtin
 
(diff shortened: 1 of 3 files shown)


apitools/base/py/batch.py
```diff
@@ -180,11 +181,14 @@ def Add(self, service, method, request, global_params=None):
             http_request, self.retryable_codes, service, method_config)
         self.api_requests.append(api_request)
 
-    def Execute(self, http, sleep_between_polls=5, max_retries=5):
+    def Execute(self, http, max_batch_size=None,
```
craigcitro: Small thing: I'd say add the new argument at the end of the list, if you want to be friendly to potential callers passing all args positionally.

Issue #139, opened 2017-01-18
Warn instead of fail on GCE cache file access errors


thobrla:
[Here](https://github.com/google/apitools/blob/cce7bd1896ba611186b0a3ce6b579da0687f158e/apitools/base/py/credentials_lib.py#L294) and [here](https://github.com/google/apitools/blob/cce7bd1896ba611186b0a3ce6b579da0687f158e/apitools/base/py/credentials_lib.py#L323) apitools attempts to interact with a cache file for GCE credentials.  Exceptions accessing the file propagate out.  Instead, apitools should only warn, since the cache file is not essential for functioning.

PR #140, opened 2017-01-18
Treat exceptions accessing GCE credential cache file as a cache miss

thobrla:
This fixes an issue where problems accessing the cache file on
the filesystem (for example, in a container with no mounted
filesystem) would cause apitools to abort entirely, as opposed to
simply not caching the credentials.

Fixes https://github.com/google/apitools/issues/139

### DIFF
--- a/apitools/base/py/credentials_lib.py
+++ b/apitools/base/py/credentials_lib.py
@@ -302,6 +302,8 @@ class GceAssertionCredentials(gce.AppAssertionCredentials):
                             if (creds['scopes'] in
                                     (None, cached_creds['scopes'])):
                                 scopes = cached_creds['scopes']
+                except KeyboardInterrupt:
+                    raise
                 except:  # pylint: disable=bare-except
                     # Treat exceptions as a cache miss.
                     pass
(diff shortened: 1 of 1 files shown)


apitools/base/py/credentials_lib.py
```diff
@@ -302,6 +302,9 @@ def _CheckCacheFileForMatch(self, cache_filename, scopes):
                             if (creds['scopes'] in
                                     (None, cached_creds['scopes'])):
                                 scopes = cached_creds['scopes']
+                except:  # pylint: disable=bare-except
```
cherba: Is it not possible to list specific exceptions we care to skip. This would also supress Ctrl-C.

apitools/base/py/credentials_lib.py
```diff
@@ -331,6 +334,9 @@ def _WriteCacheFile(self, cache_filename, scopes):
                         # If it's not locked, the locking process will
                         # write the same data to the file, so just
                         # continue.
+                except:  # pylint: disable=bare-except
```
cherba: Do we care to add any unit tests?

thobrla: It would require mocking locked_file which was removed from more recent versions of oauth2client; this code is simple enough that I'd prefer not to unless you object.

apitools/base/py/credentials_lib.py
thobrla: Sure, do you have any other suggestions beyond KeyboardInterrupt?

Issue #141, opened 2017-01-29
Global logging level changes from INFO to ERROR when apitools is invoked using a ThreadPool


chamikaramj:
This happens in Apache Beam Python SDK when trying to estimate size of a GCS file pattern using a thread pool.

This seems to be due to global logger set/reset at following location.
https://github.com/google/apitools/blob/master/apitools/base/py/encoding.py#L282

When invoking this using a ThreadPool, some threads might see 'logging.ERROR' to be the old_level and reset global logger to 'logging.ERROR'.

PR #142, opened 2017-01-29
Updates encoding.py to not to set/reset global logger.

chamikaramj:
Updates encoding.py to not to set/reset global logger.

Removes a warning log entry that caused above to be added.


chamikaramj:
This fixes #141 

@cherba please let me know how to proceed here.

chamikaramj:
Thanks.

PR #145, opened 2017-02-27
Add batch request callback hook

jcoatgoogle:
This allows custom error handling etc in the case of batch requests, similar to what is possible in single request mode.

### DIFF
--- a/apitools/base/py/batch_test.py
+++ b/apitools/base/py/batch_test.py
@@ -152,12 +152,13 @@ class BatchTest(unittest2.TestCase):
     def testSingleRequestInBatch(self):
         desired_url = 'https://www.example.com'
 
-        callback_result = None
+        callback_was_called = []
         def _Callback(response, exception):
             self.assertEqual({'status': '200'}, response.info)
             self.assertEqual('content', response.content)
             self.assertEqual(desired_url, response.request_url)
             self.assertIsNone(exception)
+            callback_was_called.append(1)
 
         mock_service = FakeService()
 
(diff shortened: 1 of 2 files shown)


jcoatgoogle:
@vilasj, I augmented a test to use this hook. I'm not sure if you want something more focused, or if it should be used more broadly, etc. I welcome any advice on how to best test this. Thanks!

apitools/base/py/batch_test.py
```diff
@@ -185,7 +193,8 @@ def testSingleRequestInBatch(self):
                 'desired_request': desired_request,
             })
 
-            api_request_responses = batch_api_request.Execute(FakeHttp())
+            api_request_responses = batch_api_request.Execute(
+                FakeHttp(), batch_request_callback=_Callback)
```
vilasj: This will pass if the callback is never executed?

jcoatgoogle: Ah drat, you're right. I've updated the test. Meant to add that originally...

apitools/base/py/batch_test.py
```diff
@@ -150,9 +150,18 @@ def testRequestServiceUnavailable(self):
                                   exceptions.HttpError)
 
     def testSingleRequestInBatch(self):
+        desired_url = 'https://www.example.com'
+
+        callback_was_called = []
```
vilasj: Minor: Just curious, why list instead of just an integer count?

jcoatgoogle: Python weirdness about callbacks and mutable variables. Calling a function on a variable will work 100% of the time, otherwise I need to mess with the global keyword.

vilasj: Ah, thanks for clarification.

vilasj:
Looks like there are some lint failures...

jcoatgoogle:
Yeah I saw the lint failures. I'll fix and resubmit.

jcoatgoogle:
@vilasj fixed the lint error.

PR #146, opened 2017-03-02
Update from 0.5.7 to 0.5.8

jcoatgoogle:


apitools/base/py/batch_test.py
```diff
@@ -153,6 +153,7 @@ def testSingleRequestInBatch(self):
         desired_url = 'https://www.example.com'
 
         callback_was_called = []
+
```
vilasj: Just for readability?

jcoatgoogle: Somehow, when I committed it it didn't make pep8 unhappy, but now it did.

PR #150, opened 2017-04-03
Fix a bug with ADC being None.

craigcitro:
Unsurprisingly, None doesn't have a `create_scoped` method. :)


vilasj:
LGTM. Could you add a small test for this?

PR #152, opened 2017-04-19
Fix new lint warnings/errors.

cherba:


vilasj:
Coverage went down....wonder why?

apitools/base/protorpclite/messages.py
```diff
@@ -250,8 +251,7 @@ class MyMessage(Message):
         outer_definition_name = cls.outer_definition_name()
         if outer_definition_name is None:
             return six.text_type(cls.__name__)
-        else:
-            return u'%s.%s' % (outer_definition_name, cls.__name__)
+        return u'%s.%s' % (outer_definition_name, cls.__name__)
```
vilasj: Curious, why is this considered better i.e. why was the previous code resulting in a lint error?

cherba: It is subjective, but having return at the end irrespective of condition is less error prone, eg if/elif/elif/else each doing return, if you forget else then it would silently return nothing.

vilasj: Thanks for the clarification.

cherba: Regarding coverage. It is a fraction of # of lines covered / # total lines.
If you subtract 1 from the numerator and denumerator the value of the fraction will go down. If it went up that would mean we removed some previously non-covered code, so it would be more reason for concern. So coverage going up is not always good thing.

All the changes in this PR should be noop.

svalentin: I approve of the linter. The idea is you want to return early from functions and not have nested if-elses that are hard to follow. This also means converting "if A return; elseif B return; else return;" to "if A return; if B return; return;". Notice the lack of elses. It makes each individual condition stand on its own without linking it to something above. Imagine having to break condition B into B1 and B2 and things get messy fast in the first example.
Some more info on this from other people - https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement

vilasj: Yup, agree that early return is more readable/maintainable. This particular instance seemed neither here nor there, but it is the base case of the pattern :)

PR #156, opened 2017-04-25
Remove extra dependencies on setuptools

sb2nov:


sb2nov:
R: @cherba PTAL. 

I think we can get rid of these dependencies all together which makes things a bit simpler.

PR #159, opened 2017-04-28
Add function to encode AdditionalProperty messages

znewman01:
It takes in a dict-like object, optionally sorts the items, and returns
a populated message.


znewman01:
@cherba PTAL

PR #163, opened 2017-05-12
Allow key of AdditionalProperty to contain unicode

houglum:
Use case: for custom object metadata, gsutil uses an AdditionalProperty
message. Custom object metadata can contain unicode characters in either
the key or value fields. Previously, the key field was being implicitly
encoded as ASCII via a call to str(). When this call fails, we should
attempt to properly decode the field. See GoogleCloudPlatform/gsutil#431
for additional context.


houglum:
@cherba PTAL (assuming you're the right person to review this)

craigcitro:
Test? 😉

houglum:
Done - edited one of the existing tests to use a unicode string for the key.

Issue #164, opened 2017-06-15
list_pager.YieldFromList should prevent batchSize > limit


pcostell:
If a limit is provided, YieldFromList should enforce that the requested batch size is less than the remaining limit. Otherwise, more data will be requested from the service than requested by the user.

This should apply to the limit as it decreases, for example:
Request 1: limit: 90, batch_size: 50 --> request with batch_size: 50
Request 2: limit: 40, batch_size: 50 --> request with batch_size: 40

PR #165, opened 2017-06-15
Request the minimum of limit and batch_size.

pcostell:
Addresses #164 by setting the batch_size on each request to the min of the limit and the batch_size.

### DIFF
--- a/apitools/base/py/list_pager.py
+++ b/apitools/base/py/list_pager.py
@@ -59,10 +59,11 @@ def YieldFromList(
 
     """
     request = encoding.CopyProtoMessage(request)
-    if batch_size_attribute:
-        setattr(request, batch_size_attribute, batch_size)
     setattr(request, current_token_attribute, None)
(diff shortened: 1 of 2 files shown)


apitools/base/py/list_pager.py
```diff
@@ -59,10 +59,12 @@ def YieldFromList(
 
     """
     request = encoding.CopyProtoMessage(request)
-    if batch_size_attribute:
-        setattr(request, batch_size_attribute, batch_size)
     setattr(request, current_token_attribute, None)
     while limit is None or limit:
+        if batch_size_attribute:
+            request_batch_size = min(batch_size,
```
vilasj: request_batch_size = min(batch_size, limit or batch_size) ?

pcostell:
The server side only receives a single "page_size". In the case of our API, we just noticed this due to some bad test data further down in the list. For some APIs, List costs money per result returned, so I believe this is good in general.

PR #166, opened 2017-08-08
Handle map<string, bytes> field encoding to base64.

cherba:


apitools/base/py/encoding_test.py
```diff
@@ -106,6 +106,20 @@ class AdditionalProperty(messages.Message):
         'AdditionalProperty', 1, repeated=True)
 
 
+class Secret(messages.Message):
```
ImJasonH: This name might be confusing without the specific context of the original bug. Hard to decipher a year from now.

cherba: Done.

PR #168, opened 2017-08-10
Make JSON mappings use the class as a dictionary key

joelzed:
This had been using a string identifier which fails when a package is
specified.

### DIFF
--- a/apitools/base/py/encoding.py
+++ b/apitools/base/py/encoding.py
@@ -659,7 +659,8 @@ def _CheckForExistingMappings(mapping_type, message_type,
 
 
 def _EncodeCustomFieldNames(message, encoded_value):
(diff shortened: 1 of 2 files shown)


apitools/base/py/encoding.py
```diff
@@ -553,30 +551,8 @@ def _ProcessUnknownMessages(message, encoded_message):
 _JSON_FIELD_MAPPINGS = {}
 
 
-def _GetTypeKey(message_type, package):
-    """Get the prefix for this message type in mapping dicts."""
-    key = message_type.definition_name()
-    if package and key.startswith(package + '.'):
-        module_name = message_type.__module__
-        # We normalize '__main__' to something unique, if possible.
-        if module_name == '__main__':
```
craigcitro: FWIW, the main use case for this that I recall was this section (where we load these files from a generated CLI, and possibly hit name mismatches).

PR #169, opened 2017-08-11
Fix encoding issues of unknown fields.

cherba:
This reverts previous pull request and moves encoding of these fields outside protorpc lib.


craigcitro:
Travis failures seem independent; note that gsutil seems to have dropped support for python 2.6 in the docs, so I think we can also drop it here.

PR #179, opened 2017-10-01
Add subclasses for HttpError for simplified control flow

### LINKED ISSUE #178: Instantiate known HTTP status codes as subclasses of HttpError
This is an FR to use separate classes for known status codes, which all may derive from HttpError, for instance:

```
class ConflictError(HttpError):
   """This is an HTTP Conflict 409 error."""
```

Why? The reason is to greatly simplify control flow for the caller. It is very common to want to switch on specifically the status code. For instance, let's say you want to display a resource, but if it doesn't yet exist you want to prompt for its creation. The current recommended flow is:

```
try:
  my_resource = MyMessageClass.Get()  # calls API
except apitools_exceptions.HttpError as e:
  if e.status_code == 409:
    my_resource = OfferCreateInteractively()
  elif e.status_code == 403:
    raise PermissionError('Please go to your console and grant access to the user account')
  else:
    raise UserFacingCatchAllError(e)
```

With real classes, the control flow could look like this:
```
try:
  my_resource = MyMessageClass.Get()  # calls API
except apitools_exceptions.HttpConflictError:
  my_resource = OfferCreateInteractively()
except apitools_exceptions.HttpForbiddenError:
  raise MyPermissionError('Please go to your console and grant access to the user account')
except apitools_exceptions.HttpError as e:
  raise UserFacingCatchAllError(e)
```

This way, we:

- avoid boilerplate for re-raising
- utilize Python-idiomatic control flow for exceptions
- allow for easier mocking of errors in unit tests

while still being fully backwards compatible.

betamos:
Closes #178

### DIFF
--- a/apitools/base/py/base_api_test.py
+++ b/apitools/base/py/base_api_test.py
@@ -198,10 +198,9 @@ class BaseApiTest(unittest2.TestCase):
         service = FakeService(client=client)
         request = SimpleMessage()
         with mock(base_api.http_wrapper, 'MakeRequest', fakeMakeRequest):
-            with self.assertRaises(exceptions.HttpError) as error_context:
+            with self.assertRaises(exceptions.HttpBadRequestError) as err:
                 service._RunMethod(method_config, request)
-        http_error = error_context.exception
-        self.assertEquals(400, http_error.status_code)
+        http_error = err.exception
         self.assertEquals('http://www.google.com', http_error.url)
         self.assertEquals('{"field": "abc"}', http_error.content)
         self.assertEquals(method_config, http_error.method_config)
--- a/apitools/base/py/exceptions.py
+++ b/apitools/base/py/exceptions.py
@@ -74,9 +74,9 @@ class HttpError(CommunicationError):
         return int(self.response['status'])
 
     @classmethod
-    def FromResponse(cls, http_response):
+    def FromResponse(cls, http_response, **kwargs):
         return cls(http_response.info, http_response.content,
-                   http_response.request_url)
+                   http_response.request_url, **kwargs)
 
 
 class InvalidUserInputError(InvalidDataError):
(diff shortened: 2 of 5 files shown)


cherba:
LGTM.

apitools/base/py/base_api_test.py
```diff
@@ -198,9 +198,9 @@ def fakeMakeRequest(*unused_args, **unused_kwargs):
         service = FakeService(client=client)
         request = SimpleMessage()
         with mock(base_api.http_wrapper, 'MakeRequest', fakeMakeRequest):
-            with self.assertRaises(exceptions.HttpError) as error_context:
+            with self.assertRaises(exceptions.HttpBadRequestError) as err:
                 service._RunMethod(method_config, request)
-        http_error = error_context.exception
+        http_error = err.exception
         self.assertEquals(400, http_error.status_code)
```
craigcitro: You can drop this line now, right? :wink:

apitools/base/py/exceptions.py
```diff
@@ -74,9 +74,48 @@ def status_code(self):
...
+            status_code = int(http_response.info.get('status'))
+            error_cls = _HTTP_ERRORS.get(status_code, cls)
+        except ValueError:
+            error_cls = cls
+        return error_cls(http_response.info, http_response.content,
+                         http_response.request_url, method_config, request)
+
+
+class HttpBadRequestError(HttpError):
+
```
craigcitro: I'd drop the newlines between class definition and docstring.

apitools/base/py/exceptions.py
```diff
@@ -143,14 +182,17 @@ class RetryAfterError(HttpError):
...
-        super(RetryAfterError, self).__init__(response, content, url)
+    def __init__(self, response, content, url, retry_after,
+                 method_config=None, request=None):
+        super(RetryAfterError, self).__init__(
+            response, content, url, method_config, request)
         self.retry_after = int(retry_after)
 
     @classmethod
-    def FromResponse(cls, http_response):
+    def FromResponse(cls, http_response, method_config=None, request=None):
```
craigcitro: Why not just `**kwargs`?

apitools/base/py/exceptions.py
```diff
@@ -143,14 +182,17 @@ class RetryAfterError(HttpError):
...
         return cls(http_response.info, http_response.content,
-                   http_response.request_url, http_response.retry_after)
+                   http_response.request_url, http_response.retry_after,
+                   method_config, request)
```
craigcitro: I'd use explicit names for the last two args (though my previous comment will make this one irrelevant).

apitools/base/py/base_api_test.py
betamos: Done!

apitools/base/py/exceptions.py
betamos: Done! (I originally kept them for consistency with the rest of the file, but I changed it now)

apitools/base/py/exceptions.py
betamos: Added **kwargs to all of the methods instead, to have single source of truth for keyword arguments in HttpError constructor. Much better, thanks.

apitools/base/py/exceptions.py
betamos: No longer relevant.

PR #182, opened 2017-10-30
Add compression support

mooman219:
Add basic gzip transport encoding support to chunked streaming uploads.

### DIFF
--- a/apitools/base/py/compression.py
+++ b/apitools/base/py/compression.py
@@ -123,7 +123,7 @@ class StreamingBuffer(object):
         if size is None:
             size = self.__size
         ret_list = []
-        while size > 0 and len(self.__buf):
+        while size > 0 and self.__buf:
             data = self.__buf.popleft()
             size -= len(data)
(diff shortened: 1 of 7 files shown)


vilasj:
Thanks for the pull request. Tests seem to be failing in Travis. Please take a look.

mooman219:
Oh darn, compatibility issues with python 3. I'll check it out

mooman219:
Gzip.py is a _mostly_ as is port from CPython 3.4 to work in 2.7/3+. It fixes a streaming bug that exists in python 2.7 and is required for streams that don't support tell() and seek(). I don't see a point if fixing the lints as I would rather keep minimal diffs.

houglum:
LGTM - I agree with keeping the source that was ported from CPython as-is (as much as possible) to enable easier merging in the future, if needed.

houglum:
Hmm... GitHub didn't give me the option to squash the 10 commits into one.  I'll pull in the repo, fix it, and force push.

PR #183, opened 2017-11-07
Add simple upload compression support.

mooman219:
This pull request adds additional support for compressing "simple" uploads. This includes multipart and media uploads. Additionally, compressing uploads when auto_transfer is set is added.


mooman219:
I really dislike this linter's opinion on how long with statements should be formatted.

Issue #162, opened 2017-05-11
Is there chance update to accommodate oauth2client>4.0.0  


sandeepkoduri:
https://github.com/cloudendpoints/endpoints-management-python/issues/21

sandeepkoduri:
Any update on this.?

sandeepkoduri:
from oauth2client import locked_file
ImportError: cannot import name locked_file

we have upgraded oauth2client an year ago, this library is still using an year old version of that package. This dependencies are holding us back for migrating to endpoints v2, for which the V1 is already deprecated. It would be great if dependencies from Google managed libraries are compatible with other Google libraries.

fili:
It actually also impacts endpoints and apitools for another reason:

```
/apitools/base/py/credentials_lib.py", line 71, in <module>
    from oauth2client import multistore_file
ImportError: cannot import name multistore_file
```

Could only find this solution (which is to downgrade): https://stackoverflow.com/a/44266031

PR #184, opened 2017-11-15
Update client to accommodate oauth2client>=4.0

e4p:
This changes has been needed for a while now. The main blocker
seems to be the use of locked_file for caching GCE credentials.
I've added a simple multiprocess lockable file cache that uses
a similar approach to that used in ouath2client's multiprocess
file storage.

Submission of this should close issue #162.

### DIFF
--- a/apitools/base/py/credentials_lib_test.py
+++ b/apitools/base/py/credentials_lib_test.py
@@ -13,6 +13,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import os.path
+import shutil
+import tempfile
+
 import mock
 import six
 import unittest2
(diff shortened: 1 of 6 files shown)


apitools/base/py/credentials_lib_test.py
```diff
@@ -21,43 +25,88 @@
...
         return credentials
 
     def testGceServiceAccounts(self):
         scopes = ['scope1']
-        self._GetServiceCreds()
-        self._GetServiceCreds(scopes=scopes)
-        self._GetServiceCreds(service_account_name='my_service_account',
-                              scopes=scopes)
+        self._GetServiceCreds(service_account_name=None,
+                                      scopes=None)
```
houglum: Did the linter not complain about the odd spacing here (and the statement below)?

e4p: It did, I've added a fix in the latest test. How about I squash this all

houglum: Sounds good. That's the only nit I had -- LGTM if you're addressing that :)

e4p:
I ran the gsutil integration tests, both on a clean clone of gsutil, and on a clone where I substituted my apitools repo into the third_party directory. The change introduced no new failures (although 2 unit tests and one integration test seemed to be already failing in the unaltered repo).

houglum:
Just ran the integration tests against:

- A fresh clone of gsutil on an Ubunutu machine, removing third_party/apitools and copying in e4p's fork into its place -- all tests passed.
- The same code, but run sequentially on a Windows 7 machine -- all passed there as well.

PR #187, opened 2017-11-20
App Engine: add thread-only fallback to process locks.

e4p:


apitools/base/py/credentials_lib.py
```diff
@@ -557,6 +571,11 @@ class _MultiProcessCacheFile(object):
     * The interprocess lock will not deadlock. If a process can not acquire
       the interprocess lock within `_lock_timeout` the call will return as
       a cache miss or unsuccessful cache write.
+    * App Engine environments cannot be process locked because (1) the runtime
+      does not provide monotonic time and (2) different processes are not
+      guaranteed to share the same machine. Because of this, process locks
```
craigcitro: guaranteed not to?

apitools/base/py/credentials_lib.py
```diff
@@ -37,6 +36,21 @@
 from apitools.base.py import exceptions
 from apitools.base.py import util
 
+# App Engine does not support ctypes which are required for the
+# monotonic time used in fasteners. Conversely, App Engine does
+# not support colocated concurrent processes, so process locks
+# are not needed.
+try:
+    import fasteners
+    _ENV_IS_APP_ENGINE = False
```
craigcitro: I think you could also try inspecting `os.environ['SERVER_SOFTWARE']`, but I'm not sure it's worth it.

WDYT of just making this `_FASTENERS_AVAILABLE` and then ignoring locking otherwise? Anyone who's `pip install`'d will have it anyway.

e4p: Sounds good.

apitools/base/py/credentials_lib.py
e4p: Resolved wording via DM

apitools/base/py/credentials_lib.py
```diff
@@ -42,14 +42,14 @@
...
-    _ENV_IS_APP_ENGINE = False
-except ImportError as fasteners_import_error:
-    try:
-        # pylint: disable=unused-import
-        import google.appengine.api.app_identity as unused_import
-    except ImportError:
-        raise fasteners_import_error
-    _ENV_IS_APP_ENGINE = True
+    _FASTENERS_AVAILABLE = True
+    raise ImportError('this is a fake import')
```
craigcitro: I'm guessing you want to drop this line? 😉

PR #188, opened 2017-11-28
Fix issue with copying unrecognized Enum values

znewman01:
Before, this resulted in an error on decoding; now, it succeeds.

I removed the comment about deleting the CopyProtoMessage function since
(1) there's uses of the function both internally and by users of
apitools, and (2) it's clearly non-trivial to correctly copy a proto
message.

### DIFF
--- a/apitools/base/py/encoding.py
+++ b/apitools/base/py/encoding.py
@@ -88,10 +88,9 @@ def RegisterFieldTypeCodec(encoder, decoder):
     return Register
 
 
-# TODO(craigcitro): Delete this function with the switch to proto2.
 def CopyProtoMessage(message):
-    codec = protojson.ProtoJson()
-    return codec.decode_message(type(message), codec.encode_message(message))
+    """Make a deep copy of a message."""
+    return JsonToMessage(type(message), MessageToJson(message))
 
 
 def MessageToJson(message, include_fields=None):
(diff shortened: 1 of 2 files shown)


znewman01:
Please let me know if I'm missing some context/history @craigcitro

znewman01:
@markpell FYI

craigcitro:
I suspect I'm the one lacking context here -- what spurred this? 😉 

@markpell do you have context to review this?

znewman01:
Ahh—so there was a bug where gcloud was using this function and a new enum value got added server-side. apitools decoded it fine, but trying to CopyProtoMessage the message with the unknown enum value resulted in a crash. This fixes that (the test case with UnknownEnum crashed before this change).

Mark knows about that crash and can verify that this change will prevent similar crashes in the future, but I wanted to know if there was any reason that CopyProtoMessage was implemented in this way originally. If you have no recollection, and all the tests pass (I checked the gcloud test suite as well), it's probably safe to move forward. Just wanted to give you veto power if I was missing something.

znewman01:
TYVM! I'll wait for @markpell  to look too.

(New revision is just a fix-up for a lint error—I'm used to 2-spaces :) )

znewman01:
Okay, great! I need someone with write access to actually merge it in now.

PR #189, opened 2017-12-06
Fix issue where uploading files would hang if a request failed while …

kevinli7:
…using stream_slice.StreamSlice.

### DIFF
--- a/apitools/base/py/transfer_test.py
+++ b/apitools/base/py/transfer_test.py
@@ -17,6 +17,7 @@
 """Tests for transfer.py."""
 import string
 
+import httplib2
 import mock
 import six
 from six.moves import http_client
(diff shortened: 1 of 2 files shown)


vilasj:
There are some lint failures.

apitools/base/py/transfer_test.py
```diff
@@ -477,3 +484,77 @@ def testMediaCompressed(self):
...
+                self.response,  # Initial request in InitializeUpload().
+                self.fail_response,  # 503 status code from server.
+                self.response # Successful request.
+            ])
+
+            # Initialization.
+            upload.InitializeUpload(self.request, bytes_http)
+            upload.StreamInChunks()
+
+            # Ensure the mock was called three times
```
vilasj: Why three? A short comment may be helpful.

Also, would it be useful to assert anything else?

kevinli7: Do you think another comment is necessary here. In the side_effect above, I've commented on the what the responses are.

vilasj: Ah ok. May be extract responses into a local variable and use the len of that in the assert?

kevinli7: Done.

PR #195, opened 2017-12-12
Fix GCE creds serialization w/ oauth2client >= 4.0

houglum:
When updating gsutil to use oauth2client 4.1.2 and apitools 0.5.19,
testing on GCE instances showed that GceAssertionCredentials were not
being written to our credential storage file. Debugging showed that this
was because of a KeyError when looking for the 'scope' attribute in a
serialized credential (no longer included as of oauth2client 3.0). The
'scope' value was also mistakenly being passed in as an argument for
'email'. This change maintains backward compatibility for oauth2client
versions that used the 'scope' attribute while also working for current
versions that don't include 'scope' when constructing a
GceAssertionCredentials object.

### DIFF
--- a/apitools/base/py/credentials_lib.py
+++ b/apitools/base/py/credentials_lib.py
@@ -268,7 +268,7 @@ class GceAssertionCredentials(gce.AppAssertionCredentials):
         # catch and squelch the warning.
         with warnings.catch_warnings():
             warnings.simplefilter('ignore')
-            super(GceAssertionCredentials, self).__init__(scopes, **kwds)
+            super(GceAssertionCredentials, self).__init__(scope=scopes, **kwds)
 
     @classmethod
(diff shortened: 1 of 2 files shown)


houglum:
FYI: Fixed the linter errors and ensured credential serialization still works on GCE.

thobrla:
Is it possible emulate each version (serializable/unserializable) of GceAssertionCredentials in a unit test to ensure to_json/from_json both work?

houglum:
Not sure how/why to test a version that doesn't support `to_json()`, but I've added a unit test for calling `to_json()` on `GceAssertionCredentials`.

thobrla:
LGTM pending fixing the Travis CI lint errors.

houglum:
D'oh.  Fixed.

PR #198, opened 2017-12-23
Fix issue where resumable uploads weren't resuming correctly on 500s.

kevinli7:
This commit does the following:
- Creates an Upload specific CheckResponse function that does not raise exception.
    - The built-in CheckResponse in http_wrapper was triggering a retry loop by raising an exception so Upload.RefreshResumableUploadState was never being called which led to uploads failing when a 500 was returned from the server. 
- Moved logic that refreshes upload state to __StreamMedia which is responsible for completing the upload.
- Fixed small issue where if a chunk wasn't uploaded completely, the upload stream would be seeked to the wrong position.

### DIFF
--- a/apitools/base/py/transfer.py
+++ b/apitools/base/py/transfer.py
@@ -165,12 +165,12 @@ class _Transfer(object):
     def EnsureInitialized(self):
         if not self.initialized:
             raise exceptions.TransferInvalidError(
-                'Cannot use uninitialized %s', self._type_name)
+                'Cannot use uninitialized %s' % self._type_name)
 
     def EnsureUninitialized(self):
         if self.initialized:
             raise exceptions.TransferInvalidError(
-                'Cannot re-initialize %s', self._type_name)
+                'Cannot re-initialize %s' % self._type_name)
 
     def __del__(self):
         if self.__close_stream:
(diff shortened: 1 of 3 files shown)


kevinli7:
@houglum 
I've run gsutil's tests with this change and it seems to work fine. If you could confirm that would be great!

apitools/base/py/transfer.py
```diff
@@ -913,7 +918,22 @@ def CallSendMediaBody(start):
             if response.status_code in (http_client.OK, http_client.CREATED):
                 self.__complete = True
                 break
-            self.__progress = self.__GetLastByte(response.info['range'])
+            if response.status_code not in (
+                    http_client.OK, http_client.CREATED,
+                    http_wrapper.RESUME_INCOMPLETE):
+                # We want to reset our state to wherever the server left us
+                # before this failed request, and then raise.
+                self.RefreshResumableUploadState()
```
houglum: Nit: Can you move the call to RefreshResumableUploadState() down past line 933, so that it only takes place after we know the strategy is "RESUMABLE_UPLOAD"?  I know RefreshResumableUploadState() will return if it's not a resumable upload, but it would look semantically cleaner if we only called it once we knew the upload was a resumable one :)

kevinli7: Done!

kevinli7:
Moved the call to RefreshResumableUploadState.

PR #202, opened 2018-02-28
Fix issue with encoding unrecognized Enum values.

khtg:
Previously this caused an error when the Enum values were inside
an unknown field because the base protojson codec used to copy the message
during the mapping of unknown fields fails on unknown Enums. This changes
the base protojson to handle unknown values for EnumFields as
unrecognized fields.

### DIFF
--- a/apitools/base/protorpclite/protojson.py
+++ b/apitools/base/protorpclite/protojson.py
@@ -283,11 +283,19 @@ class ProtoJson(object):
                 valid_value = [self.decode_field(field, item)
                                for item in value]
                 setattr(message, field.name, valid_value)
-            else:
-                # This is just for consistency with the old behavior.
-                if value == []:
-                    continue
+                continue
+            # This is just for consistency with the old behavior.
+            if value == []:
+                continue
+            try:
                 setattr(message, field.name, self.decode_field(field, value))
+            except messages.DecodeError:
+                # Save unknown enum values.
+                if not isinstance(field, messages.EnumField):
+                    raise
+                variant = self.__find_variant(value)
+                if variant:
+                    message.set_unrecognized_field(key, value, variant)
 
         return message
 
--- a/apitools/base/protorpclite/protojson_test.py
+++ b/apitools/base/protorpclite/protojson_test.py
@@ -198,12 +198,18 @@ class ProtojsonTest(test_util.TestCase,
 
     def testNumericEnumerationNegativeTest(self):
         """Test with an invalid number for the enum value."""
-        self.assertRaisesRegexp(
-            messages.DecodeError,
-            'Invalid enum value "89"',
-            protojson.decode_message,
-            MyMessage,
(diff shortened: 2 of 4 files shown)


khtg:
Confirmed that gcloud tests work with this change.
@houglum I've also run the gsutil tests with this change as we discussed yesterday and they all passed.

apitools/base/protorpclite/test_util.py
```diff
@@ -579,11 +579,17 @@ def testContentType(self):
...
-                                         self.encoded_invalid_enum)
+        # Since protos need to be able to add new enums, a message should be
+        # successfully decoded even if the enum value is invalid. Encoding the
+        # decoded message should result in equivalence with the original
+        # encoded message containing an invalid enum.
+        decoded = self.PROTOLIB.decode_message(
+            OptionalMessage, self.encoded_invalid_enum)
+        message = OptionalMessage()
+        self.assertEqual(message, decoded)
+        encoded = self.PROTOLIB.encode_message(
```
houglum: Nit: This line break seems silly

khtg: No, sorry, just copy-pasting without reformatting I think. Went through and condensed a bit, please take a look.

apitools/base/protorpclite/protojson.py
```diff
@@ -283,11 +283,19 @@ def __decode_dictionary(self, message_type, dictionary):
...
+                continue
+            # This is just for consistency with the old behavior.
+            if value == []:
+                continue
+            try:
                 setattr(message, field.name, self.decode_field(field, value))
+            except messages.DecodeError:
+                # Save unknown enum values.
+                if not isinstance(field, messages.EnumField):
+                    raise
```
vilasj: When is this going to happen? Is raising the decode error sufficient in that situation?

apitools/base/protorpclite/protojson_test.py
```diff
@@ -198,12 +198,16 @@ def testNumericEnumeration(self):
 
     def testNumericEnumerationNegativeTest(self):
         """Test with an invalid number for the enum value."""
-        self.assertRaisesRegexp(
```
vilasj: Add other tests for the DecodeError that will still be raised?

apitools/base/protorpclite/protojson.py
khtg: This happens in decode_field for misformatted datetime fields, base64 encoding errors, and message fields that can't be decoded recursively. I believe this is sufficient because those all reflect actual problems with the encoded message.

apitools/base/protorpclite/protojson_test.py
khtg: Done. testDecodeInvalidDateTime and testDecodeBadBase64BytesField already did this, added testDecodeInvalidDateTime for a MessageField.

PR #203, opened 2018-03-07
Fixing crash on Python 3 when batch_size is set to None in the list pager

markpell:
It's unclear what the behavior of passing None should actually do (I think the current behavior is wrong) but either way it just crashes on Python 3. The change in this PR makes Py3 behave the way Py2 currently does.

### DIFF
--- a/apitools/base/py/list_pager.py
+++ b/apitools/base/py/list_pager.py
@@ -62,7 +62,16 @@ def YieldFromList(
     setattr(request, current_token_attribute, None)
     while limit is None or limit:
         if batch_size_attribute:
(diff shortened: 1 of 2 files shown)


apitools/base/py/list_pager.py
```diff
@@ -62,7 +62,16 @@ def YieldFromList(
     setattr(request, current_token_attribute, None)
     while limit is None or limit:
         if batch_size_attribute:
-            request_batch_size = min(batch_size, limit or batch_size)
+            # On Py3, None is not compariable so min() below will fail.
```
vilasj: typo: comparable

PR #211, opened 2018-03-29
exclude samples & tests from install

vapier:
There's no value in installing the sample modules (plus they pollute
the global samples.* namespace), nor in installing the unittests.

For the samples, we set the include option with find_packages so the
code only searches the top level apitools dir.

For the tests, because they live alongside modules vs a completely
sep tree (which is fine), we have to add a series of globs to the
exclude_package_data to filter them out.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -68,7 +68,7 @@ setuptools.setup(
     author='Craig Citro',
     author_email='craigcitro@google.com',
     # Contained modules and scripts.
-    packages=setuptools.find_packages(),
+    packages=setuptools.find_packages(include=['apitools']),
     entry_points={'console_scripts': CONSOLE_SCRIPTS},
     install_requires=REQUIRED_PACKAGES,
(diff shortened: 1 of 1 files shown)


craigcitro:
From the sidelines: 👍 on dropping these bits on install.

kevinli7:
Does this exclude samples? 

Also, how does one verify this works? I tried running `python setup.py install` and the tests were not excluded.

vapier:
> Does this exclude samples?

yes.  see the change to the `packages` line.  that tells setuptools to only look in the `apitools/` top level dir which means the `samples/` top level dir is completely ignored.  imo, this is what we want as samples are for devs who are working on the package (akin to documentation) rather than something that installed packages care about.

> Also, how does one verify this works? I tried running python setup.py install and the tests were not excluded.

it's working for me ...
```sh
$ rm -rf build foo/ && python ./setup.py build && python ./setup.py install --root $PWD/foo
...
$ tree foo/
...nothing related to tests in here...
```

if i revert my patch and re-run that snippet, then foo/ is full of test files.

PR #213, opened 2018-04-04
Try lint on Python 3 instead of Python 2

cclauss:
Python 3 is more strict on syntax issues than Python 2


cclauss:
Closed on favor of #251

PR #219, opened 2018-04-17
Use UTF-8 instead of ASCII for the StringField validator encoding

jakeakopp:
Use UTF-8 instead of ASCII for the StringField validator encoding.

### DIFF
--- a/apitools/base/protorpclite/messages_test.py
+++ b/apitools/base/protorpclite/messages_test.py
@@ -616,7 +616,7 @@ class FieldTest(test_util.TestCase):
             messages.InvalidDefaultError,
             r"Invalid default value for StringField:.*: "
             r"Field encountered non-UTF-8 string .*: "
-            r"'utf.*8' codec can't decode byte 0xc3 in position 0: "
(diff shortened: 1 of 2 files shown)


apitools/base/protorpclite/messages_test.py
```diff
@@ -615,10 +615,10 @@ def testStringField_BadUnicodeInDefault(self):
         self.assertRaisesWithRegexpMatch(
             messages.InvalidDefaultError,
             r"Invalid default value for StringField:.*: "
-            r"Field encountered non-ASCII string .*: "
-            r"'ascii' codec can't decode byte 0x89 in position 0: "
-            r"ordinal not in range",
-            messages.StringField, 1, default=b'\x89')
+            r"Field encountered non-UTF-8 string .*: "
+            r"'utf.*8' codec can't decode byte 0xc3 in position 0: "
+            r"invalid continuation byte",
```
kevinli7: nit: would utf-?8 be better here. I assume different python versions are doing utf8 vs utf-8? This would be more explicit, and .* seems to be overkill for this.

PR #226, opened 2018-06-01
python3 modernize open() => io.open() and convert comment unicode homoglyphs 

cococlyde:
python3 modernize open() => io.open() and convert comment unicode homoglyphs  to their ASCII equivalents.


cococlyde:
I signed it!

kevinli7:
Ignoring travis failure as it's a failure of install wheel on python 3.3

PR #230, opened 2018-06-13
Define long() in Python 3

cclauss:
__long()__ was remove in Python 3 in favor of __int()__.


tvalentyn:
I don't have a strong opinion, but FYI feature detection is also acceptable and recommended in https://docs.python.org/3/howto/pyporting.html#use-feature-detection-instead-of-version-detection

kevinli7:
Thanks for the reference. That's fine then.

kevinli7:
Thanks for the reference. That's fine then.

PR #231, opened 2018-06-13
Builtin cmp() was removed in Python 3

cclauss:
http://portingguide.readthedocs.io/en/latest/comparisons.html#the-cmp-function

### DIFF
--- a/apitools/base/protorpclite/messages.py
+++ b/apitools/base/protorpclite/messages.py
@@ -49,6 +49,12 @@ import six
 
 from apitools.base.protorpclite import util
 
+try:
+  cmp             # Python 2
+except NameError:
+  def cmp(x, y):  # Python 3
+    return (x > y) - (x < y)
+
 __all__ = [
     'MAX_ENUM_VALUE',
     'MAX_FIELD_NUMBER',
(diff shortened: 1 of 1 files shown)


apitools/base/protorpclite/messages.py
```diff
@@ -476,12 +476,6 @@ def __reduce__(self):
         """
         return self.__class__, (self.number,)
 
-    def __cmp__(self, other):
```
tvalentyn: Do we actually need to remove __cmp__? I think it will just be ignored on Python 3, and looks like this class already implements rich comparison methods.

cclauss: Python 3 moved to rich comparisons for speed (evaluate each element once instead of evaluating each element multiple times) so given that Python 2 expires at yearend, let’s just drop the slower approach now.

tvalentyn: So, on Python 3 `__cmp__` will not be executed, but it's not clear to me how long people will use apitools on Python 2, even after expiration some users may still be stuck with Python 2. Without digging deeper into this codebase (which I am not familiar with), it seems harmless to keep this as is until apitools drops support for Python 2.

cclauss: https://docs.python.org/3.7/howto/sorting.html#the-old-way-using-the-cmp-parameter

Issue #232, opened 2018-06-14
Importing from apitools.base.py submodules is slow


HarrisonGregg:
`apitools/base/py/__init__.py` imports everything from all of its submodules. This makes it easier to import things in the submodules, as you don't have to worry about which submodule its in, but it comes at the price of significant overhead when importing anything from any of the submodules, as all of the submodules (and everything they import) have to be imported too. On my machine, this takes about 160ms. These imports should be removed as wildcard imports in an `__init__.py` file are generally bad practice and have specific consequences in this case.

PR #234, opened 2018-06-14
Reorganize encoding.py so it doesn't rely on imports in __init__.py

HarrisonGregg:
Prerequisite for addressing #232.

Currently, when you import `apitools.base.py.encoding`, it triggers the imports in `apitools/base/py/__init__.py`. This includes importing `extra_types.py`, which makes patches to `encoding`, such as:
```python
encoding.RegisterCustomMessageCodec(
    encoder=JsonProtoEncoder, decoder=_JsonToJsonValue)(JsonValue)
```
That means that if you remove the imports in `__init__.py`, then `encoding` behaves differently because `extra_types` isn't imported. This change reorganizes `encoding.py`, so that whenever it is imported, `extra_types.py` is also imported.

### DIFF
--- a/apitools/base/py/encoding.py
+++ b/apitools/base/py/encoding.py
@@ -1,0 +1,38 @@
+#!/usr/bin/env python
+#
+# Copyright 2015 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


vilasj:
Travis build is failing: https://travis-ci.org/google/apitools/builds/392441331?utm_source=github_status&utm_medium=notification.

PR #237, opened 2018-06-15
Include 201 CREATED in HTTP success response codes

sixolet:
Previously, we only accepted 200 OK and 204 NO_CONTENT. Some APIs we'd like to use this library with return 201 CREATED when creating a resource, so that should also count as success (and behave like 200 OK).


vilasj:
Lint checks are failing on travis: https://travis-ci.org/google/apitools/builds/392829610?utm_source=github_status&utm_medium=notification.

PR #238, opened 2018-06-20
Catch another invalid \escape.

cococlyde:
Catch another \escape that crashed the JSON reader. This one somehow originated from &nbsp; in the proto text.


cococlyde:
I didn't realize commentary is raw HTML. The invalid escape somehow originated from <ampersand>nbsp<semicolon> in the proto text.

PR #240, opened 2018-08-07
Clean service names in the registry so as not to write syntax errors

sixolet:
Generating a client with a dash in the service name produced invalid code.


kevinli7:
Travis CI is failing due to unrelated changes.

PR #244, opened 2018-09-13
 Adding type checking of method request and response mocks

markpell:
Currently these mocks are not checked to ensure that the messages being expected and returned could actually occur. It is generally invalid to use types that don't match, but if you want to, you can set enable_type_checking=False on your expect statement.

### DIFF
--- a/apitools/base/py/testing/mock.py
+++ b/apitools/base/py/testing/mock.py
@@ -191,8 +191,14 @@ class _MockedMethod(object):
         self.__mocked_client = mocked_client
         self.__real_method = real_method
         self.method_config = real_method.method_config
+        config = self.method_config()
+        self.__request_type = getattr(self.__mocked_client.MESSAGES_MODULE,
+                                      config.request_type_name)
+        self.__response_type = getattr(self.__mocked_client.MESSAGES_MODULE,
+                                       config.response_type_name)
 
-    def Expect(self, request, response=None, exception=None, **unused_kwargs):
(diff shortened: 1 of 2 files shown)


apitools/base/py/testing/mock.py
```diff
@@ -191,8 +191,39 @@ def __init__(self, key, mocked_client, real_method):
...
+        """Ensure the given message is of the expected type of this method.
+
+        Args:
+          msg: The message instance to check.
+          is_request: True to validate against the expected request type,
+             False to validate against the expected response type.
+
+        Raises:
+          exceptions.ConfigurationValueError: If the type of the message was
+             not correctly.
```
kevinli7: not correct?

apitools/base/py/testing/mock.py
```diff
@@ -191,8 +191,39 @@ def __init__(self, key, mocked_client, real_method):
...
+            real_type = self.__response_type
+
+        if not isinstance(msg, real_type):
+            raise exceptions.ConfigurationValueError(
+                'Expected {} is not of the correct type for this method.\n'
+                '   Required: [{}]\n'
+                '   Given:    [{}]'.format(mode, real_type, type(msg)))
+
+    def Expect(self, request, response=None, exception=None,
+               enable_type_checking=True, **unused_kwargs):
```
kevinli7: Are we having this as an option because there are a lot of cases where this fails?

vilasj:
Nice change, scary that this was not checked before!

Issue #246, opened 2018-10-05
TravisCI tests began failing at some commit after v0.5.24


houglum:
TravisCI tests began failing at some commit after v0.5.24 was released.  Here is a link to an example set of CI tests; looks like it fails under the lint and Py3.4+ configurations:
https://travis-ci.org/google/apitools/builds/437401234?utm_source=github_status&utm_medium=notification

From the commit history, it seems that https://github.com/google/apitools/commit/8c502250841d5d4cccb24a18403bdcb980b1b383 was the culprit -- I believe the commit was pushed directly to this repo, rather than going through the pull request flow, which would have run the CI tests and caught this.

houglum:
@markpell, can you take a look at why the new mock type checking functionality is failing under the above-mentioned CI runs?

PR #247, opened 2018-10-11
Fixing tests

markpell:
I'm not sure why this test only fails on py3, but this is what the test should actually be like.

### DIFF
--- a/apitools/base/py/testing/mock_test.py
+++ b/apitools/base/py/testing/mock_test.py
@@ -46,8 +46,8 @@ class MockTest(unittest2.TestCase):
 
     def testMockFusionBasic(self):
         with mock.Client(fusiontables.FusiontablesV1) as client_class:
-            client_class.column.List.Expect(request=1, response=2,
-                enable_type_checking=False)
+            client_class.column.List.Expect(
+                request=1, response=2, enable_type_checking=False)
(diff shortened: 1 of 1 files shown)


markpell:
Fixes #246

markpell:
I've never been able to get CI to run on my PRs. Any idea why?

houglum:
I can get this to work when I push a commit to a branch on my fork... all I can think of that might be helpful is:

-  If you don't have a TravisCI account set up, do that (https://docs.travis-ci.com/user/getting-started/)
-  Then, if you go to https://travis-ci.org/markpell/apitools/settings, make sure that the "Build pushes branches" and "Build pushed pull requests" toggle-switches are enabled.

I have those settings enabled and was able to CI to run when I added your fork as a remote, fetched from it, checked out your `textfix` branch, and pushed those changes to my branch here:
https://github.com/houglum/apitools/commits/test-textfix

FYI, it looks like the Py3 tests are passing now, but there were some lint errors.  I fixed them in this commit, if you'd like to pull in that commit from my fork into yours:
https://github.com/houglum/apitools/commit/200fb56c2252059f6924e87bc293b4d70233a828

markpell:
Thanks! Yeah it looks like the travis config was just messed up. Fixed the lint errors and all presubmits now pass.

Issue #249, opened 2018-12-04
Python 3 incompatible in batch request


markflyhigh:
Batch request failed in Python 3 using apitools library due to response content encoding error. Found this problem in Beam Python pipeline whose gcsio is built based on apitools. The version I use is python 3.5.

Stacktrace:
File "/usr/local/lib/python3.5/site-packages/apache_beam/io/iobase.py", line 1077, in <genexpr>
    window.TimestampedValue(v, timestamp.MAX_TIMESTAMP) for v in outputs)
  File "/usr/local/lib/python3.5/site-packages/apache_beam/io/filebasedsink.py", line 315, in finalize_write
    num_threads)
  File "/usr/local/lib/python3.5/site-packages/apache_beam/internal/util.py", line 145, in run_using_threadpool
    return pool.map(fn_to_execute, inputs)
  File "/usr/local/lib/python3.5/multiprocessing/pool.py", line 266, in map
    return self._map_async(func, iterable, mapstar, chunksize).get()
  File "/usr/local/lib/python3.5/multiprocessing/pool.py", line 644, in get
    raise self._value
  File "/usr/local/lib/python3.5/multiprocessing/pool.py", line 119, in worker
    result = (True, func(*args, **kwds))
  File "/usr/local/lib/python3.5/multiprocessing/pool.py", line 44, in mapstar
    return list(map(*args))
  File "/usr/local/lib/python3.5/site-packages/apache_beam/io/filebasedsink.py", line 299, in _rename_batch
    FileSystems.rename(source_files, destination_files)
  File "/usr/local/lib/python3.5/site-packages/apache_beam/io/filesystems.py", line 252, in rename
    return filesystem.rename(source_file_names, destination_file_names)
  File "/usr/local/lib/python3.5/site-packages/apache_beam/io/gcp/gcsfilesystem.py", line 229, in rename
    copy_statuses = gcsio.GcsIO().copy_batch(batch)
  File "/usr/local/lib/python3.5/site-packages/apache_beam/io/gcp/gcsio.py", line 322, in copy_batch
    api_calls = batch_request.Execute(self.client._http)  # pylint: disable=protected-access
  File "/usr/local/lib/python3.5/site-packages/apitools/base/py/batch.py", line 222, in Execute
    batch_http_request.Execute(http)
  File "/usr/local/lib/python3.5/site-packages/apitools/base/py/batch.py", line 480, in Execute
    self._Execute(http)
[...]

PR #250, opened 2018-12-20
Handle encoded response content in batch http request in Python 3

markflyhigh:
Fix issue reported in https://github.com/google/apitools/issues/249.

The problem happens in Python 3 when execute a http request and encoded content is returned. This breaks `email_parser.Parser().parsestr()` call in `batch.py` since the function argument only accepted `str` (not bytes). There is a callback `__ProcessHttpResponse` that handles encoded response content but the call happens later.

### DIFF
--- a/apitools/base/py/base_api.py
+++ b/apitools/base/py/base_api.py
@@ -610,7 +610,7 @@ class BaseApiService(object):
                 request_url=http_response.request_url)
 
         content = http_response.content
-        if self._client.response_encoding:
+        if self._client.response_encoding and isinstance(content, bytes):
             content = content.decode(self._client.response_encoding)
 
         if self.__client.response_type_model == 'json':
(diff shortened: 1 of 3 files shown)


markflyhigh:
friendly ping.

kevinli7:
When I test this out manually in gcloud, this line in batch.py is failing in python2:
mime_response = parser.parsestr(header + content)

Also, I feel like this decoding should have some sort of switch to turn it on and off.

markflyhigh:
Interesting. The unit tests are passing in Python 2 and 3. Can you give more details how you test with gcloud?

Add a switch to this decoding sounds good to me. I think we can add a `response_encoding` field to `BatchHttpRequest` and `BatchApiRequest`, which also allow people to specify the encoding of the response content.

markflyhigh:
@kevinli7 PTAL

kevinli7:
LGTM. @houglum not sure if you guys use batch? Do you want to take a look as well?

PR #252, opened 2019-01-22
Use print() function in both Python 2 and Python 3

cclauss:
Like #228  Legacy __print__ statements are syntax errors in Python 3 but __print()__ function works as expected in both Python 2 and Python 3.

### DIFF
--- a/ez_setup.py
+++ b/ez_setup.py
@@ -60,9 +60,9 @@ def _validate_md5(egg_name, data):
         digest = md5(data).hexdigest()
         if digest != md5_data[egg_name]:
             print(
-                "md5 validation of %s failed!  (Possible download problem?)"
-                % egg_name
-            , file=sys.stderr)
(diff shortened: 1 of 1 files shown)


ez_setup.py
```diff
@@ -93,12 +94,12 @@ def do_download():
         pkg_resources.require("setuptools>="+version); return
     except pkg_resources.VersionConflict as e:
         if was_imported:
-            print >>sys.stderr, (
+            print((
```
houglum: Nit: There are unnecessary parentheses around the format string now; the string is implicitly continued via being in the opening parens for `print(`.

ez_setup.py
```diff
@@ -204,10 +205,10 @@ def main(argv, version=DEFAULT_VERSION):
                 os.unlink(egg)
     else:
         if setuptools.__version__ == '0.0.1':
-            print >>sys.stderr, (
+            print((
```
houglum: Same comment as above.

PR #255, opened 2019-02-13
Merge in ehodapp's Python 3 compat changes

catleeball:
These were leftover changes from Ed's VM that Joe and I build on top of. This is the first commit the in the cherry-picking process of merging in Joe and my own changes to help with Python 2.7 and 3.x compatibility. Joe has since rebuilt some of these components such as ttywrite.

### DIFF
--- a/apitools/base/py/base_api.py
+++ b/apitools/base/py/base_api.py
@@ -529,20 +529,51 @@ class BaseApiService(object):
         return query_info
 
     def __FinalUrlValue(self, value, field):
-        """Encode value for the URL, using field to skip encoding for bytes."""
+        """Returns the appropriate "string" type for the given field.
+
+        For all field types except `messages.BytesField`:
+        For PY2 values, we want bytes; for PY3, we want unicode. Regardless of
+        language version, all returned values will be able to be encoded
+        appropriately to be sent over the wire later. If passed bytes that
+        cannot be represented with the utf-8 encoding, this method will raise
+        an exception when trying to convert them to unicode.
+
+        Note regarding BytesField fields:
+        The "field" parameter, if it's given and is an instance of
+        `messages.BytesField`, will result in using a B64-encoded representation
+        of the bytes. If in PY3, these bytes will be decoded to unicode.
+
+        This method returns a value of type `six.text_type` (bytes for PY2,
+        unicode for PY3).
+        """
+
+        # Otherwise, we want to return the language version's native type for
+        # "strings" (unicode in PY3, bytes, encoded such that they're valid for
+        # use in URLs, in PY3).
         if isinstance(field, messages.BytesField) and value is not None:
-            return base64.urlsafe_b64encode(value)
-        elif isinstance(value, datetime.datetime):
-            return value.isoformat()
-        if six.PY3:
-            if isinstance(value, bytes):
+            value = base64.urlsafe_b64encode(value)
+            if six.PY2:
+                return value
+            else:  # Python >= 3:
+                # Base64 charset can be represented using ascii/utf-8.
                 return value.decode('utf-8')
-            return value
-        else:
+        if isinstance(value, datetime.datetime):
+            # The isoformat() method returns the language's "string" type.
+            return value.isoformat()
+        if six.PY2:
             if isinstance(value, six.text_type):
                 return value.encode('utf8')
-            elif isinstance(value, six.binary_type):
-                return value.decode('utf8')
+            if isinstance(value, six.binary_type):
+                # First ensure that bytes can be decoded as utf-8. This ensures
+                # that we don't pass along bytes that are invalid for use with
+                # URLs, e.g. latin-1 encoded bytes.
+                _ = value.decode('utf-8')  # Throws an exception if invalid.
+                return value
+        else:  # Python >= 3
+            if isinstance(value, six.text_type):
+                return value
+            if isinstance(value, six.binary_type):
+                return value.decode('utf-8')
         return value
 
     def __ConstructQueryParams(self, query_params, request, global_params):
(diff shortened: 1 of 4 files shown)


apitools/base/py/base_api.py
```diff
@@ -614,10 +614,11 @@ def __ProcessHttpResponse(self, method_config, http_response, request):
             content = content.decode(self._client.response_encoding)
 
         if self.__client.response_type_model == 'json':
-            return content
+            return http_response.content
```
houglum: I *think* this should go back to being `content` (since it could have had its type changed a couple lines up when it was decoded).  Same for line 621 below.

apitools/base/py/transfer.py
```diff
@@ -336,12 +336,10 @@ def __NormalizeStartEnd(self, start, end=None):
         if end is not None:
             if start < 0:
                 raise exceptions.TransferInvalidError(
-                    'Cannot have end index with negative start index ' +
-                    '[start=%d, end=%d]' % (start, end))
+                    'Cannot have end index with negative start index')
```
houglum: Same; Ed's commits are old enough that they didn't pull these changes in.

apitools/base/py/transfer.py
```diff
@@ -749,15 +767,9 @@ def ConfigureRequest(self, upload_config, http_request, url_builder):
...
-                # Turn the body into a stream so that we can compress it, then
-                # read the compressed bytes.  In the event of a retry (e.g. if
-                # our access token has expired), we need to be able to re-read
-                # the body, which we can't do with a stream. So, we consume the
-                # bytes from the stream now and store them in a re-readable
-                # bytes container.
-                http_request.body = (
-                    compression.CompressStream(
-                        six.BytesIO(http_request.body))[0].read())
+                body_buffer = six.BytesIO(http_request.body)
```
houglum: Same; Ed's commits are old enough that they didn't pull these changes in.

apitools/base/py/base_api.py
catleeball: Whoops, I'll fix this! So preferring Ed's version, right?

apitools/base/py/base_api.py
```diff
@@ -536,12 +532,17 @@ def __FinalUrlValue(self, value, field):
         """Encode value for the URL, using field to skip encoding for bytes."""
         if isinstance(field, messages.BytesField) and value is not None:
             return base64.urlsafe_b64encode(value)
-        elif isinstance(value, six.text_type):
-            return value.encode('utf8')
-        elif isinstance(value, six.binary_type):
-            return value.decode('utf8')
         elif isinstance(value, datetime.datetime):
             return value.isoformat()
+        if six.PY3:
```
houglum: Apparently my long comment got left out of my last batch of comments. So, I think the intent of this function is to get a "string" type native to the language version.  The `value.decode('utf-8')` bit in the old code block was just there to make sure that the bytes were valid UTF-8, and the rest of the code relied on the unicode being implicitly converted back to bytes later.  Given that understanding, I suggest we better document the function, e.g.:

```python
        """Returns the appropriate "string" type for the given field.

        For all field types except `messages.BytesField`:
        For PY2 values, we want bytes; for PY3, we want unicode. Regardless of
        language version, all returned values will be able to be encoded
        appropriately to be sent over the wire later. If passed bytes that
        cannot be represented with the utf-8 encoding, this method will raise
        an exception when trying to convert them to unicode.

        Note regarding BytesField fields:
        The "field" parameter, if it's given and is an instance of
        `messages.BytesField`, will result in using a B64-encoded representation
        of the bytes. If in PY3, these bytes will be decoded to unicode.

        This method returns a value of type `six.text_type` (bytes for PY2,
        unicode for PY3).
        """

        # Otherwise, we want to return the language version's native type for
        # "strings" (unicode in PY3, bytes, encoded such that they're valid for
        # use in URLs, in PY3).

        if isinstance(field, messages.BytesField) and value is not None:
            value = base64.urlsafe_b64encode(value)
            if six.PY2:
                return value
            else:  # Python >= 3:
                # Base64 charset can be represented using ascii/utf-8.
                return value.decode('utf-8')
        if isinstance(value, datetime.datetime):
            # The isoformat() method returns the language's "string" type.
            return value.isoformat()
        if six.PY2:
            if isinstance(value, six.text_type):
                return value.encode('utf8')
            if isinstance(value, six.binary_type):
                # First ensure that bytes can be decoded as utf-8. This ensures
                # that we don't pass along bytes that are invalid for use with
                # URLs, e.g. latin-1 encoded bytes.
                _ = value.decode('utf-8')  # Throws an exception if invalid.
                return value
        else:  # Python >= 3
            if isinstance(value, six.text_type):
                return value
            if isinstance(value, six.binary_type):
                return value.decode('utf-8')
        return value
```

catleeball: Got it! That docstring looks good to me, I'll add it in.

houglum: Note that I tweaked most of the function as well to make the control flow a bit more clear and explain why each step was necessary.  Let me know if you still find it hard to follow, or if you think the previous version was more clear.

apitools/base/py/transfer.py
```diff
@@ -483,13 +487,6 @@ def GetRange(self, start, end=None, additional_headers=None,
             response = self.__ProcessResponse(response)
             progress += response.length
             if response.length == 0:
-                if response.status_code == http_client.OK:
```
catleeball: I'm not super sure why this was removed. It feels like it should be okay regardless of version? Even if http_client.OK doesn't exist in some version, it would resolve to None and fall through to the raise exception below right? Maybe this was here for some debugging purpose?

apitools/base/py/encoding.py
```diff
@@ -307,7 +307,8 @@ def decode_field(self, field, value):
             value = result.value
             if result.complete:
                 return value
-        if isinstance(field, messages.MessageField):
+        if isinstance(field, messages.MessageField) and not isinstance(
```
catleeball: Thanks for adding this fix, great find! Are there any other classes that inherit from Message other than DateTimeField we should watch for here?

catleeball:
Closing this PR after talking with @walkerjoe , we're just going to grab a couple relevant methods from Ed's old code instead of merging it in wholesale.

PR #259, opened 2019-02-25
Support GZip compression when fetching discovery URL

qdii:


apitools/gen/util.py
```diff
@@ -373,7 +388,16 @@ def FetchDiscoveryDoc(discovery_url, retries=5):
...
             try:
-                content = urllib_request.urlopen(url).read()
+                response = urllib_request.urlopen(url)
+                encoding = response.info().get('Content-Encoding')
+                if encoding == 'gzip':
+                  content = _Gunzip(response.read())
+                elif encoding == 'deflate':
+                  content = response.read()
+                else:
+                  logging.warning('Unknown encoding')
```
jameslynnwu: We shouldn't log a warning if the encoding is not specified. `Content-Encoding` looks like an optional header.

https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.11

apitools/gen/util.py
```diff
@@ -373,7 +388,16 @@ def FetchDiscoveryDoc(discovery_url, retries=5):
     for url in discovery_urls:
         for _ in range(retries):
             try:
-                content = urllib_request.urlopen(url).read()
+                response = urllib_request.urlopen(url)
```
jameslynnwu: Can we pull the request code into a method and add tests for the possible content encodings? 

Something similar to https://github.com/httplib2/httplib2/blob/master/tests/test_encoding.py

I was surprised that deflate is handled the same way as no encoding.

apitools/gen/util.py
qdii: Done.

apitools/gen/util.py
qdii: Done.

qdii:
Looks like my last commit broke compatibility with python2, working on a patch…

qdii:
Should be good now.

qdii:
Fixed the lint.

apitools/gen/util.py
```diff
@@ -365,6 +367,30 @@ def _NormalizeDiscoveryUrls(discovery_url):
...
+        os.unlink(f.name)
+
+
+def _GetURLContent(url):
+    """Download and return the content of URL."""
+    response = urllib_request.urlopen(url)
+    encoding = response.info().get('Content-Encoding')
+    if encoding == 'gzip':
+        content = _Gunzip(response.read())
+    else:
```
jameslynnwu: would be nice to also add deflate support :)

qdii: IIUC, it also works with deflate: I just removed the branch as it just repeated what the default branch did. Or am I missing something?

jameslynnwu: I expected some code like, https://github.com/httplib2/httplib2/blob/master/python2/httplib2/__init__.py#L521

PR #262, opened 2019-03-03
Add __hash__ to class Message to make it hashable in Python 3

ttanay:
The class Message defines a __eq__ method, but, does not define a
__hash__ method. This makes it unhashable in Python 3. But, it is
hashable in Python 2.

### DIFF
--- a/apitools/base/protorpclite/messages.py
+++ b/apitools/base/protorpclite/messages.py
@@ -758,6 +758,8 @@ class Message(six.with_metaclass(_MessageClass, object)):
 
     """
 
+    __hash__ = object.__hash__
+
     def __init__(self, **kwargs):
         """Initialize internal messages state.
 


apitools/base/protorpclite/messages.py
```diff
@@ -758,6 +758,8 @@ class Order(Message):
 
     """
 
+    __hash__ = object.__hash__
```
tvalentyn: I think hash should be consistent with implementation of `__eq__`, so that equal messages have equal hashes.

ttanay:
Hi @tvalentyn 
Thanks for reviewing.

I've made the change. PTAL.

apitools/base/protorpclite/messages.py
```diff
@@ -1060,6 +1058,10 @@ def __ne__(self, other):
         """
         return not self.__eq__(other)
 
+    def __hash__(self):
+        """Hash by __tags"""
+        return hash(self.__tags)
```
tvalentyn: Looking closer at this, `self.__tags` is a dictionary, which is unhashable. Looks like we have two options:
- Keep the class unhashable. 
- Hash over contents of `__tags` or it's subset, if we can consider Message objects immutable, which we probably cannot.

Is there a reason to believe Message objects are immutable? If not, I think making them unhashable makes most sense.

ttanay: `self.__tags` is a dict! I missed that, sorry :man_facepalming: 

I think Message should be mutable. Therefore, it should be unhashable.
This class is hashable in Python 2. Making it unhashable would break compatibility in Python 3. But, that is also an option. 

If we make it unhashable, we should add `__hash__ = None` to the class to keep the behaviour consistent in Python 2 as well.
OR
If we want to keep behaviour consistent in Python 3, while not breaking compatibility, we should add `__hash__ = object.__hash__` to the class.

tvalentyn: There are good reasons why in Python 3 `__hash__` is not implemented by default, in this light, I think the better way to keep compatibility would be to set `__hash__ = None`. This may require cascading changes for Py2 users of `apitools`. But same/similar changes would still be needed in downstream codebase to make relevant code Python3-compatible, so that may encourage users of apitools to fix their code sooner than later.

ttanay: Cool. I'll make that change :)

ttanay:
@tvalentyn Made the change :rocket: 
PTAL :)

tvalentyn:
Looks good to me.

tvalentyn:
Looping in @houglum who is a committer in this repo -  PTAL, thank you.

houglum:
Sorry for the delay, didn't see this until late.

I'll pull this change into our submodule of apitools in gsutil tomorrow, then run all the integration tests to make sure we weren't relying on this.  Please do @ me if I forget :)

tvalentyn:
Thank you @houglum . No worries :)

ttanay:
Can this be merged?

PR #263, opened 2019-03-18
Python 2 / 3 compatibility changes

catleeball:
@walkerjoe and I made several small fixes to help apitools be cross-compatible with Python 2 and 3. These issues were identified while working on making [gsutil's py-six-current branch](https://github.com/GoogleCloudPlatform/gsutil/tree/py-six-current) compatible with Python 2.7 and 3.5+. This was done under the guidance of @houglum 

Notably, we're using the [six](https://github.com/benjaminp/six) library to help ensure byte strings vs unicode strings work together nicely.

Email generator had some small text encoding issues with generating text in Python 3.5+, so we added `MultipartBytesGenerator` method to alleviate the issue.

Let us know if you have any questions about these changes! :slightly_smiling_face:

### DIFF
--- a/apitools/base/protorpclite/protojson.py
+++ b/apitools/base/protorpclite/protojson.py
@@ -206,6 +206,7 @@ class ProtoJson(object):
           ValueError: If encoded_message is not valid JSON.
           messages.ValidationError if merged message is not initialized.
         """
+        encoded_message = six.ensure_str(encoded_message)
         if not encoded_message.strip():
             return message_type()
 
--- a/apitools/base/py/transfer.py
+++ b/apitools/base/py/transfer.py
@@ -426,7 +426,7 @@ class Download(_Transfer):
                 raise exceptions.TransferRetryError(response.content)
         if response.status_code in (http_client.OK,
                                     http_client.PARTIAL_CONTENT):
-            self.stream.write(response.content)
+            self.stream.write(six.ensure_binary(response.content))
             self.__progress += response.length
             if response.info and 'content-encoding' in response.info:
                 # TODO(craigcitro): Handle the case where this changes over a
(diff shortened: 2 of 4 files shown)


catleeball:
Note that both Joe and I have signed CLAs, but googlebot gets a little upset with multiple authors. :writing_hand:

apitools/base/py/transfer.py
```diff
@@ -541,6 +544,25 @@ def StreamMedia(self, callback=None, finish_callback=None,
         self._ExecuteCallback(finish_callback, response)
 
 
+if six.PY3:
+    class MultipartBytesGenerator(email_generator.BytesGenerator):
+        """Generates a bytes version of a Message object tree for multipart messages
```
houglum: Linter will probably complain that this line is more than 80 chars long.  Also needs an extra line of whitespace between the summary line and the description paragraph.

apitools/base/protorpclite/protojson.py
```diff
@@ -206,6 +206,7 @@ def decode_message(self, message_type, encoded_message):
           ValueError: If encoded_message is not valid JSON.
           messages.ValidationError if merged message is not initialized.
         """
+        encoded_message = six.ensure_str(encoded_message)
```
houglum: We'll need to change the `setup.py` file to require a newer minimum version of `six` for this, right? (I think the `ensure_<foo>` stuff was introduced in 1.12, and we currently only require >=1.9).

apitools/base/py/transfer.py
catleeball: Good catch, fixed.

apitools/base/protorpclite/protojson.py
catleeball: You're right! Updated to >= 1.12

PR #268, opened 2019-05-17
allow sets to be used for repeated fields

mikedanese:
Part of fixing http://issuetracker.google.com/132917477

```
Field scope is repeated. Found: set([u'https://www.googleapis.com/auth/compute', u'https://www.googleapis.com/auth/appengine.admin', u'https://www.googleapis.com/auth/userinfo.email', u'https://www.googleapis.com/auth/cloud-platform'])
```

Disclaimer: no idea what's going on with apitools or python, just want the bug fixed. Feel free to co-opt this change.

### DIFF
--- a/apitools/base/protorpclite/messages.py
+++ b/apitools/base/protorpclite/messages.py
@@ -1358,7 +1358,7 @@ class Field(six.with_metaclass(_FieldMeta, object)):
             return validate_element(value)
         else:
             # Must be a list or tuple, may not be a string.
-            if isinstance(value, (list, tuple)):
+            if isinstance(value, (list, tuple, set)):
                 result = []
                 for element in value:
                     if element is None:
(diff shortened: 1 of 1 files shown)


apitools/base/protorpclite/messages.py
```diff
@@ -1358,7 +1358,7 @@ def __validate(self, value, validate_element):
             return validate_element(value)
         else:
             # Must be a list or tuple, may not be a string.
-            if isinstance(value, (list, tuple)):
+            if isinstance(value, (list, tuple, set)):
```
mikedanese: This is line fixed the issue for me, but I added the second for good measure.

jameslynnwu:
I'm not sure if this is the right approach. It might be better to make sure the caller is always converting sets into lists before assigning it to the field. What do you think?

mikedanese:
Can you give some examples of downsides of accepting sets here? I don't have enough perspective to weigh the pros and cons. Sets don't have a better fit in proto than a repeated field.

jameslynnwu:
Just the fact that currently repeated fields have the property that order matters, whether intended or not. I'd rather not change that.

mikedanese:
Thanks! That makes sense to me.

PR #269, opened 2019-06-04
Ensure unicode strings for Python 2 / 3 compat

catleeball:
Made several changes to `credentials_lib.py` for Python 2/3 compatibility. Specifically, Running pre-release tests, we found some of these methods would fail when gsutil had no configs available.

### DIFF
--- a/apitools/base/py/credentials_lib.py
+++ b/apitools/base/py/credentials_lib.py
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 """Common credentials classes and constructors."""
+from __future__ import unicode_literals
 from __future__ import print_function
 
 import contextlib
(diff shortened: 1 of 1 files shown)


apitools/base/py/credentials_lib.py
```diff
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 """Common credentials classes and constructors."""
+from __future__ import unicode_literals
```
houglum: I'd rather not import this in just one file; I don't think most of the other files in this directory use this import.  (If we're going to use unicode literals, we should do it across the entire library.)

apitools/base/py/credentials_lib.py
```diff
@@ -352,7 +356,7 @@ def _ScopesFromMetadataServer(self, scopes):
     def GetServiceAccount(self, account):
         relative_url = 'instance/service-accounts'
         response = _GceMetadataRequest(relative_url)
-        response_lines = [line.rstrip('/\n\r')
+        response_lines = [six.ensure_text(line).rstrip('/\n\r')
```
houglum: If we're going to remove the `unicode_literals` import, we should make sure we're stripping off unicode here (i.e. adding the `u` prefix: `rstrip(u'/\n\r')`)

catleeball: Makes sense to me! Made both changes. :)

catleeball:
@houglum Thanks for the approval! I don't have write access to the apitools repo, so feel free to merge this when you have a moment :slightly_smiling_face:

PR #272, opened 2019-06-21
Normalize response from the GCE metadata server.

houglum:
The response is always read in as bytes, and sometimes blindly passed to
things that accept strings. In Python 3, some things that accept strings
(like json.loads) will only accept unicode-type objects, not bytes-type
objects. This change makes sure the bytes are converted to a unicode
string (decoded using the utf-8 codec) in Python 3.

Fixes https://github.com/GoogleCloudPlatform/gsutil/issues/819.


houglum:
@catleeball for FYI.

Issue #274, opened 2019-06-26
Cloud Endpoints, python, UnboundLocalError: local variable 'scope_spec' referenced before assignment


myloginistaken:
Using App Engine with Cloud Endpoints Framework for Python. In **apitools.base.py.credentials_lib.py** at l**ine 140** the variable **scope_spec** is referenced before assignment.
Full stacktrace:
```(/base/alloc/tmpfs/dynamic_runtimes/python27g/6a5167f0cae8960d/python27/python27_lib/versions/1/google/appengine/runtime/wsgi.py:269)

Traceback (most recent call last):

  File "/base/alloc/tmpfs/dynamic_runtimes/python27g/6a5167f0cae8960d/python27/python27_lib/versions/1/google/appengine/runtime/wsgi.py", line 240, in Handle
    handler = _config_handle.add_wsgi_middleware(self._LoadHandler())

  File "/base/alloc/tmpfs/dynamic_runtimes/python27g/6a5167f0cae8960d/python27/python27_lib/versions/1/google/appengine/runtime/wsgi.py", line 311, in _LoadHandler
    handler, path, err = LoadObject(self._handler)

  File "/base/alloc/tmpfs/dynamic_runtimes/python27g/6a5167f0cae8960d/python27/python27_lib/versions/1/google/appengine/runtime/wsgi.py", line 85, in LoadObject
    obj = __import__(path[0])

  File "/base/data/home/apps/e~aimydata/20190626t111403.419174768539258548/main.py", line 167, in <module>
    api = endpoints.api_server([AIMyDataAPI])

  File "/base/data/home/apps/e~aimydata/20190626t111403.419174768539258548/lib/endpoints/apiserving.py", line 601, in api_server
    controller.start()

  File "/base/data/home/apps/e~aimydata/20190626t111403.419174768539258548/lib/endpoints_management/control/client.py", line 253, in start
    self._initialize_flushing()

  File "/base/data/home/apps/e~aimydata/20190626t111403.419174768539258548/lib/endpoints_management/control/client.py", line 372, in _initialize_flushing
    self._flush_schedule_report_aggregator()

  File "/base/data/home/apps/e~aimydata/20190626t111403.419174768539258548/lib/endpoints_management/control/client.py", line 465, in _flush_schedule_report_aggregator
    transport = self._create_transport()

  File "/base/data/home/apps/e~aimydata/20190626t111403.419174768539258548/lib/endpoints_management/control/client.py", line 151, in create_transport
[...]

PR #275, opened 2019-06-26
Fix unbound local variable 'scope_spec'

jameslynnwu:
https://github.com/google/apitools/issues/274

This validation is already done by util.NormalizeScopes and was duplicate code.


catleeball:
With this in 3.40 RC 21 and ran pre-release tests, looks good from gsutil's perspective! :) http://sponge/fbde4109-6a85-4561-954e-b5f3a4a2afcd

PR #276, opened 2019-06-26
Fix mock for urllib_request in GetURLContentTest

jameslynnwu:
We were replacing the urlopen and it wasn't being unmocked, so other tests got the value 'regular non-gzipped content' when calling urlopen.


houglum:
Context: This doesn't screw anything up *now*, since the test is torn down once those two methods are finished, but it would break things if we added another method in the future.  +1 to this change, since it removes side effects that were previously sticking around after an individual test was finished.

PR #278, opened 2019-07-03
Cast mimetype to lang-appropriate str type.

houglum:
Fixes https://issuetracker.google.com/issues/136688879. This line will
fail in Python 3 if mime_type is a `bytes` object:

    if '/' not in mime_type:

with this error:

    TypeError: a bytes-like object is required, not 'str'


houglum:
Closing this in favor of enforcing that callers pass the correct string type (bytes vs unicode) along to apitools instead.  In gsutil's case, the PR is here:
https://github.com/GoogleCloudPlatform/gsutil/pull/845

PR #281, opened 2019-09-19
Pager helper to work with nested fields for limit or continuation token

sixolet:
Allow a limit or a continuation token to be in a nested field in the list request or response.

This allows, for example, if the continuation token is `metadata.continue_` like it is for a Kubernetes-style api, where `metadata` is a `ListMeta` and common to all list responses.

This still needs a little more tests.

### DIFF
--- a/apitools/base/py/list_pager.py
+++ b/apitools/base/py/list_pager.py
@@ -22,6 +22,24 @@ __all__ = [
     'YieldFromList',
 ]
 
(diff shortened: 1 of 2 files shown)


sixolet:
Ok, here we go. Docstrings added, lint fixed, *not* rewritten as iterative as discussed out of band (it's clearer code in recursive form)

PR #282, opened 2019-09-19
Switch apitools default JSON endpoint to storage.googleapis.com

mfschwartz:
This is for GCS split domain serving. Please contact me at mfschwartz@google.com if you have questions.


houglum:
This only updates the sample, right?  The actual base URL is determined when you run apitools and fetch the API's discovery doc (e.g. for storage, see the "baseUrl" attribute near the top of the JSON document at https://www.googleapis.com/discovery/v1/apis/storage/v1/rest).

mfschwartz:
Ah, I see. In that case I'm going to revert this change, and let apitools pick up the base URL from the discovery doc (which we are updating separately, and in % stages, now).

Thanks.

PR #284, opened 2019-10-22
Fix: Avoid name check if JSON keyfile is provided

wintermelons:
In `credentials_lib.py` moved `service_account_name` and `service_account_keyfile` check after JSON check, since JSON credentials does not depend on either one to be present.


wintermelons:
@googlebot I signed it!

jameslynnwu:
This seems reasonable, but have you considered upgrading to [google-auth](https://github.com/googleapis/google-auth-library-python)?

wintermelons:
@jameslynnwu Our team's code currently still uses apitools to authenticate, and there aren't plans to migrate as long as apitools is still supported AFAIK.

PR #288, opened 2020-02-10
Client generation can handle the mtlsRootUrl

liuchaoren:


apitools/gen/util.py
```diff
@@ -174,9 +174,8 @@ def NormalizeVersion(version):
     return version.replace('.', '_')
 
 
-def _ComputePaths(package, version, discovery_doc):
-    full_path = urllib_parse.urljoin(
-        discovery_doc['rootUrl'], discovery_doc['servicePath'])
+def _ComputePaths(package, version, root_url, service_path):
```
jameslynnwu: Can we add a doc string to this? Also would be good to provide examples of what package, root_url, and service_path represent.

PR #295, opened 2020-05-07
Add compatibility with httplib2 0.16.0+

ejochman:
See more at googleapis/google-api-python-client#803.

Fix applied as recommended in https://github.com/googleapis/google-api-python-client/issues/803#issuecomment-578151576

### DIFF
--- a/apitools/base/py/http_wrapper.py
+++ b/apitools/base/py/http_wrapper.py
@@ -339,6 +339,10 @@ def MakeRequest(http, http_request, retries=7, max_retry_wait=60,
     """
     retry = 0
     first_req_time = time.time()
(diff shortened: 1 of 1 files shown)


tvalentyn:
Hi @houglum, do you think you can help review this and create a patch release?

houglum:
I've mostly moved away from doing gsutil (and associated projects') maintainence; you'll want to poke @dilipped for a review :)

dilipped:
The change looks good to me, but I have never worked on this lib before, so if one of the previous contributors can take a look as well then that would be great. @jameslynnwu I see that you have approved PRs in the past, can you please take a look at this?
Additionally, I do not have access to approve this PR and someone will have to add me to the repo

PR #304, opened 2021-04-13
Fix encoding issue with rfc822 messages

dilipped:
Addresses the issue here b/176474117

This fix works, but I am not sure if this is a bit risky given that we are using surrogateescape now.


houglum:
This seems fine. AFAICT, this only affects uncommon content types like RFC822.  (If it affected other encodings, I assume we would have seen the error mentioned in the bug by now.)

PR #305, opened 2021-04-16
Handle nano secs in DateTime parsing.

dilipped:
For context b/180215836

Fields like `custom-time` can have nano secs granularity if they are updated using non-Python libraries like the C++ client library or using the REST API directly.
```
custom_time=2020-08-12T13:39:44.51213724Z
```
Python's datetime can  handle only up to microsec precision which means a datetime string with nanosec precision would raise error.  This becomes a problem for gsutil commands like `ls` and `stat` if there are nanosec precision datetime string in the object's metadata.

Note that this change **will affect all decode_field** interaction. For example, if someone uses [JsonToMessage](https://github.com/google/apitools/blob/a48cf52831e6ef4b3ca4fd9aa95b625f6a18b8c8/apitools/base/py/encoding_helper.py#L121) which ultimately calls the `decode_field` for a datetime field, the precision will be truncated. This means that this truncation will also affect cases where a user decides to create a message from json manually with nanosec datetime precision. Hence I am adding a warning log as the best way to warn the user. 

Note that something like
```
gsutil.py setmeta -h custom-time:2020-08-12T13:39:44.512137243Z gs://bucket/obj
```
will still fail (which is good), because `setmeta` does not rely on the decode_field method.

Ideally we would want this precision truncation to be done only for display purpose and it should raise an error for all data modification cases, but this is difficult to handle with the current structure because the datetime decoding is handled
deep down the framework, and the [ProtoJson](https://github.com/google/apitools/blob/a48cf52831e6ef4b3ca4fd9aa95b625f6a18b8c8/apitools/base/protorpclite/protojson.py#L136) class is a [singleton](https://github.com/google/apitools/blob/a48cf52831e6ef4b3ca4fd9aa95b625f6a18b8c8/apitools/base/py/encoding_helper.py#L299), which means we cannot set this based on the each request/response.

I am happy to discard this PR if this seems to risky.

### DIFF
--- a/apitools/base/py/base_api.py
+++ b/apitools/base/py/base_api.py
@@ -443,7 +443,7 @@ class BaseApiClient(object):
     def DeserializeMessage(self, response_type, data):
         """Deserialize the given data as method_config.response_type."""
         try:
-            message = encoding.JsonToMessage(response_type, data)
+            message = encoding.JsonToMessage(response_type, data  )
         except (exceptions.InvalidDataFromServerError,
                 messages.ValidationError, ValueError) as e:
             raise exceptions.InvalidDataFromServerError(
--- a/apitools/base/py/base_api.py
+++ b/apitools/base/py/base_api.py
@@ -443,7 +443,7 @@ class BaseApiClient(object):
     def DeserializeMessage(self, response_type, data):
         """Deserialize the given data as method_config.response_type."""
         try:
-            message = encoding.JsonToMessage(response_type, data  )
+            message = encoding.JsonToMessage(response_type, data)
         except (exceptions.InvalidDataFromServerError,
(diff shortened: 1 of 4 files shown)


gerbercj:
LGTM

apitools/base/py/base_api.py
```diff
@@ -443,7 +443,7 @@ def SerializeMessage(self, message):
     def DeserializeMessage(self, response_type, data):
         """Deserialize the given data as method_config.response_type."""
         try:
-            message = encoding.JsonToMessage(response_type, data)
+            message = encoding.JsonToMessage(response_type, data  )
```
jameslynnwu, line 446: rm spaces

dilipped: Done.

PR #311, opened 2021-07-27
Updated messages py for unpickling related issue

aaltay:
To support migration to Python 3.7, setstate is called after append, extend, so when unpickling __field is not set. Guarding the validation to not occur during unpickle resolves this issue.

### DIFF
--- a/apitools/base/protorpclite/messages.py
+++ b/apitools/base/protorpclite/messages.py
@@ -1139,12 +1139,14 @@ class FieldList(list):
 
     def append(self, value):
         """Validate item appending to list."""
-        self.__field.validate_element(value)
+        if getattr(self, '_FieldList__field', None):
+            self.__field.validate_element(value)
         return list.append(self, value)
 
     def extend(self, sequence):
         """Validate extension of list."""
-        self.__field.validate(sequence)
+        if getattr(self, '_FieldList__field', None):
+            self.__field.validate(sequence)
         return list.extend(self, sequence)
 
     def insert(self, index, value):


aaltay:
@houglum @dilipped

aaltay:
Tests are failing for py2 with:

ERROR: Package 'rsa' requires a different Python: 2.7.15 not in '>=3.5, <4'

It does not seem related to this pr, but looks like a general config error. Tests are passing for py3 Can I ignore the failing tests?

apitools/base/protorpclite/messages.py
```diff
@@ -1139,12 +1139,14 @@ def __setitem__(self, index, value):
...
         """Validate item appending to list."""
-        self.__field.validate_element(value)
+        if getattr(self, '_FieldList__field', None):
+            self.__field.validate_element(value)
         return list.append(self, value)
 
     def extend(self, sequence):
         """Validate extension of list."""
-        self.__field.validate(sequence)
+        if getattr(self, '_FieldList__field', None):
```
dilipped, line 1148: `hasattr` seems more appropriate? https://docs.python.org/3/library/functions.html#hasattr

dilipped:
> Tests are failing for py2 with:
> 
> ERROR: Package 'rsa' requires a different Python: 2.7.15 not in '>=3.5, <4'
> 
> It does not seem related to this pr, but looks like a general config error. Tests are passing for py3 Can I ignore the failing tests?

I think we can ignore the Py2 errors.

apitools/base/protorpclite/messages.py
aaltay: Yes, it would make sense but it would result in a divergent code path with the internal version of the library. Would it be fine to keep it in this current form?

dilipped: I see. I think it's fine.

aaltay:
Thank you @dilipped. Do you know how/who could trigger a new release?

